PR #5, opened 2022-07-21
Fix `NewConfig` func & and add more unit tests

mihaelabalutoiu:
* Fix NewConfig function
  * Move the default setting before validating the config, otherwise the code will
never run. This happens because the `Validate()` function will always fail when
`config.Default.ConfigDir` is empty.
* Add unit tests for full coverage to the NewConfig function

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,9 +52,6 @@ const (
 	// DefaultConfigFilePath is the default path on disk to the garm
 	// configuration file.
 	DefaultConfigFilePath = "/etc/garm/config.toml"
-	// DefaultConfigDir is the default path on disk to the config dir. The config
-	// file will probably be in the same folder, but it is not mandatory.
-	DefaultConfigDir = "/etc/garm"
 
 	// DefaultUser is the default username that should exist on the instances.
 	DefaultUser = "runner"
@@ -73,6 +70,10 @@ const (
 )
 
 var (
+	// DefaultConfigDir is the default path on disk to the config dir. The config
+	// file will probably be in the same folder, but it is not mandatory.
+	DefaultConfigDir = "/etc/garm"
+	
 	// DefaultUserGroups are the groups the default user will be part of.
 	DefaultUserGroups = []string{
 		"sudo", "adm", "cdrom", "dialout",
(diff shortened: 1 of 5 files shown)


config/config_test.go
```diff
@@ -654,3 +656,70 @@ func TestTimeToLiveDuration(t *testing.T) {
...
+	} else {
+		require.NotNil(t, cfg)
+		defaultCfg.Default.ConfigDir = DefaultConfigDir
+		require.Equal(t, &defaultCfg, cfg)
+		require.Nil(t, err)
+	}
+}
+
+func TestNewConfigInvalidToml(t *testing.T) {
+	cfg, err := NewConfig("this is not a file path")
```
gabriel-samfira, line 710: This tests if we pass in a non existing config. The name should suggest that. Something like ```TestNewConfigInvalidTomlPath``` ?

config/config_test.go
```diff
@@ -654,3 +656,70 @@ func TestTimeToLiveDuration(t *testing.T) {
...
+	require.Nil(t, cfg)
+	require.NotNil(t, err)
+	require.Regexp(t, "decoding toml", err.Error())
+}
+
+func TestNewConfigInvalidConfig(t *testing.T) {
```
gabriel-samfira, line 716: An empty file is actually not an invalid ```toml``` config. The struct will still be populated with default values, but it will not pass ```Validate()``` because we expect that the user will set some values manually. For some options we have defaults, but some we expect to get from the user.

It would be useful to write some bogus data into that file and call ```NewConfig()``` on that. This way we test how the function will behave when the config is corrupt in some way.

config/config.go
```diff
@@ -100,12 +100,12 @@ func NewConfig(cfgFile string) (*Config, error) {
 	if _, err := toml.DecodeFile(cfgFile, &config); err != nil {
 		return nil, errors.Wrap(err, "decoding toml")
 	}
-	if err := config.Validate(); err != nil {
-		return nil, errors.Wrap(err, "validating config")
-	}
 	if config.Default.ConfigDir == "" {
 		config.Default.ConfigDir = DefaultConfigDir
 	}
+	if err := config.Validate(); err != nil {
```
gabriel-samfira, line 106: Awesome! Thanks for catching this.

gabriel-samfira (changes requested):

PR #7, opened 2022-08-17
Refactor code to allow more unit testing

mihaelabalutoiu:
In order to allow mocking for some of the `runner` functions, we created a
separate interface (called `PoolManagerController`) with `Create`, `Get`,
`Delete` operations for the `organization` / `repository` pool managers.

Furthermore, a new runner struct (`poolManagerCtrl`) implements this new
interface. The existing code is refactored to use the `poolManagerCtrl`
whenever the pool managers for `org` / `repo` are handled.

This allows more unit testing for the runner functions since `poolManagerCtrl`
field can be mocked now.

Besides this, there are some typos fixed as well.

### DIFF
--- a/runner/interfaces.go
+++ b/runner/interfaces.go
@@ -1,0 +1,34 @@
+// Copyright 2022 Cloudbase Solutions SRL
+//
+//    Licensed under the Apache License, Version 2.0 (the "License"); you may
+//    not use this file except in compliance with the License. You may obtain
+//    a copy of the License at
+//
+//         http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+//    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+//    License for the specific language governing permissions and limitations
+//    under the License.
+
+package runner
+
+import (
+	"context"
+	dbCommon "garm/database/common"
+	"garm/params"
+	"garm/runner/common"
+)
+
+type PoolManagerController interface {
+	CreateRepoPoolManager(ctx context.Context, repo params.Repository, providers map[string]common.Provider, store dbCommon.Store) (common.PoolManager, error)
+	GetRepoPoolManager(repo params.Repository) (common.PoolManager, error)
+	DeleteRepoPoolManager(repo params.Repository) error
+	GetRepoPoolManagers() (map[string]common.PoolManager, error)
+
+	CreateOrgPoolManager(ctx context.Context, org params.Organization, providers map[string]common.Provider, store dbCommon.Store) (common.PoolManager, error)
+	GetOrgPoolManager(org params.Organization) (common.PoolManager, error)
+	DeleteOrgPoolManager(org params.Organization) error
+	GetOrgPoolManagers() (map[string]common.PoolManager, error)
+}
--- a/runner/organizations.go
+++ b/runner/organizations.go
@@ -24,7 +24,6 @@ import (
 	runnerErrors "garm/errors"
 	"garm/params"
 	"garm/runner/common"
-	"garm/runner/pool"
 
 	"github.com/pkg/errors"
 )
@@ -46,7 +45,7 @@ func (r *Runner) CreateOrganization(ctx context.Context, param params.CreateOrgP
 	_, err = r.store.GetOrganization(ctx, param.Name)
 	if err != nil {
 		if !errors.Is(err, runnerErrors.ErrNotFound) {
-			return params.Organization{}, errors.Wrap(err, "fetching repo")
+			return params.Organization{}, errors.Wrap(err, "fetching org")
 		}
 	} else {
 		return params.Organization{}, runnerErrors.NewConflictError("organization %s already exists", param.Name)
@@ -63,11 +62,16 @@ func (r *Runner) CreateOrganization(ctx context.Context, param params.CreateOrgP
 		}
 	}()
 
-	poolMgr, err := r.loadOrgPoolManager(org)
-	if err := poolMgr.Start(); err != nil {
-		return params.Organization{}, errors.Wrap(err, "starting pool manager")
+	poolMgr, err := r.poolManagerCtrl.CreateOrgPoolManager(r.ctx, org, r.providers, r.store)
+	if err != nil {
+		return params.Organization{}, errors.Wrap(err, "creating org pool manager")
 	}
-	r.organizations[org.ID] = poolMgr
+	if err := poolMgr.Start(); err != nil {
+		if deleteErr := r.poolManagerCtrl.DeleteOrgPoolManager(org); deleteErr != nil {
+			log.Printf("failed to cleanup pool manager for org %s", org.ID)
+		}
+		return params.Organization{}, errors.Wrap(err, "starting org pool manager")
+	}
 	return org, nil
 }
 
@@ -91,7 +95,7 @@ func (r *Runner) GetOrganizationByID(ctx context.Context, orgID string) (params.
 
 	org, err := r.store.GetOrganizationByID(ctx, orgID)
 	if err != nil {
-		return params.Organization{}, errors.Wrap(err, "fetching repository")
+		return params.Organization{}, errors.Wrap(err, "fetching organization")
 	}
 	return org, nil
 }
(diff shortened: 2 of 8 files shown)


runner/interfaces.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	dbCommon "garm/database/common"
+	"garm/params"
+	"garm/runner/common"
+)
+
+type PoolManagerController interface {
+	CreateRepoPoolManager(ctx context.Context, repo params.Repository, providers map[string]common.Provider, store dbCommon.Store) (common.PoolManager, error)
+	GetRepoPoolManager(repo params.Repository) (common.PoolManager, error)
+	DeleteRepoPoolManager(repo params.Repository)
```
gabriel-samfira, line 27: Is ```DeleteRepoPoolManager()``` guaranteed to succeed? If not, it should at least return an ```error```, even if we end up just logging that error.

runner/interfaces.go
```diff
@@ -0,0 +1,34 @@
...
+	GetRepoPoolManagers() map[string]common.PoolManager
```
gabriel-samfira, line 28: Same as above.

runner/interfaces.go
```diff
@@ -0,0 +1,34 @@
...
+
+	CreateOrgPoolManager(ctx context.Context, org params.Organization, providers map[string]common.Provider, store dbCommon.Store) (common.PoolManager, error)
+	GetOrgPoolManager(org params.Organization) (common.PoolManager, error)
+	DeleteOrgPoolManager(org params.Organization)
```
gabriel-samfira, line 32: Idem.

runner/interfaces.go
```diff
@@ -0,0 +1,34 @@
...
+	GetOrgPoolManagers() map[string]common.PoolManager
```
gabriel-samfira, line 33: idem

runner/repositories.go
```diff
@@ -63,11 +60,20 @@ func (r *Runner) CreateRepository(ctx context.Context, param params.CreateRepoPa
...
+	}
+	repo.Internal = internalCfg
+
+	poolMgr, err := r.poolManagerCtrl.CreateRepoPoolManager(r.ctx, repo, r.providers, r.store)
+	if err != nil {
+		return params.Repository{}, errors.Wrap(err, "creating repo pool manager")
+	}
 	if err := poolMgr.Start(); err != nil {
-		return params.Repository{}, errors.Wrap(err, "starting pool manager")
+		r.poolManagerCtrl.DeleteRepoPoolManager(repo)
```
gabriel-samfira, line 74: If this fails, we need to know.

runner/repositories.go
```diff
@@ -128,6 +126,8 @@ func (r *Runner) DeleteRepository(ctx context.Context, repoID string) error {
 		return runnerErrors.NewBadRequestError("repo has pools defined (%s)", strings.Join(poolIds, ", "))
 	}
 
+	r.poolManagerCtrl.DeleteRepoPoolManager(repo)
```
gabriel-samfira, line 129: If this fails, we need to know. We may be left with dangling goroutines.

runner/organizations.go
```diff
@@ -63,11 +60,20 @@ func (r *Runner) CreateOrganization(ctx context.Context, param params.CreateOrgP
 		}
 	}()
 
-	poolMgr, err := r.loadOrgPoolManager(org)
+	internalCfg, err := r.getInternalConfig(org.CredentialsName)
```
gabriel-samfira, line 63: Calling ```getInternalConfig()``` and setting the ```Internal``` field every time before calling ```CreateOrgPoolManager()``` feels a bit tedious. The internal config is useful to the pool manager, but doesn't make sense to have it be a part of the ```Organization{}``` or ```Repository{}```. I made a mistake adding it there. If we're refactoring this bit anyway, and creating an interface for controllers, I think we should separate that info from the repo and org structs.

That information can be part of the ```PoolManager{}``` implementation or part of the ```PoolManagerController{}``` implementation. The information in the internal config is from the config file, which can't really be updated without a restart, so it makes sense to initialize it once, inject it into the pool manager controller.

This would allow us to clean up this bit of repetitive code. What do you think?

runner/organizations.go
```diff
@@ -63,11 +60,20 @@ func (r *Runner) CreateOrganization(ctx context.Context, param params.CreateOrgP
...
+	}
+	org.Internal = internalCfg
+
+	poolMgr, err := r.poolManagerCtrl.CreateOrgPoolManager(r.ctx, org, r.providers, r.store)
+	if err != nil {
+		return params.Organization{}, errors.Wrap(err, "creating org pool manager")
+	}
 	if err := poolMgr.Start(); err != nil {
-		return params.Organization{}, errors.Wrap(err, "starting pool manager")
+		r.poolManagerCtrl.DeleteOrgPoolManager(org)
```
gabriel-samfira, line 74: We need to know if this fails.

runner/organizations.go
```diff
@@ -125,9 +123,11 @@ func (r *Runner) DeleteOrganization(ctx context.Context, orgID string) error {
 			poolIds = append(poolIds, pool.ID)
 		}
 
-		return runnerErrors.NewBadRequestError("repo has pools defined (%s)", strings.Join(poolIds, ", "))
+		return runnerErrors.NewBadRequestError("org has pools defined (%s)", strings.Join(poolIds, ", "))
 	}
 
+	r.poolManagerCtrl.DeleteOrgPoolManager(org)
```
gabriel-samfira, line 129: We need to know if this fails.

runner/runner.go
```diff
@@ -252,7 +339,7 @@ func (r *Runner) Wait() error {
 
 	var wg sync.WaitGroup
 
-	for poolId, repo := range r.repositories {
+	for poolId, repo := range r.poolManagerCtrl.GetRepoPoolManagers() {
```
gabriel-samfira, line 342: ```GetRepoPoolManagers()``` should also return an error, unless it's absolutely guaranteed to never fail under any circumstances.

runner/runner.go
```diff
@@ -232,13 +319,13 @@ func (r *Runner) Stop() error {
...
 
-	for _, repo := range r.repositories {
+	for _, repo := range r.poolManagerCtrl.GetRepoPoolManagers() {
 		if err := repo.Stop(); err != nil {
 			return errors.Wrap(err, "stopping repo pool manager")
 		}
 	}
 
-	for _, org := range r.organizations {
+	for _, org := range r.poolManagerCtrl.GetOrgPoolManagers() {
```
gabriel-samfira, line 328: ```GetOrgPoolManagers()``` should also return an error, unless it's absolutely guaranteed to never fail under any circumstances.

runner/runner.go
```diff
@@ -334,9 +421,23 @@ func (r *Runner) DispatchWorkflowJob(hookTargetType, signature string, jobData [
 
 	switch HookTargetType(hookTargetType) {
 	case RepoHook:
-		poolManager, err = r.findRepoPoolManager(job.Repository.Owner.Login, job.Repository.Name)
```
gabriel-samfira, line 337: This feels more verbose than needed. Couldn't ```findRepoPoolManager()``` be refactored to use ```GetRepoPoolManager()```? You don't need to remove ```findRepoPoolManager()``` to be able to test.

gabriel-samfira (changes requested):

runner/interfaces.go
mihaelabalutoiu: In the current implementation, `DeleteRepoPoolManager()` could fail only when `PoolManager` is stopped. 

I will update the interface to have an error return value for `DeleteRepoPoolManager()`.

runner/interfaces.go
mihaelabalutoiu: I don't believe this ever fail, because it only return the repo pool managers map (which is a field in the struct that implements the interface):
```
func (p *poolManagerCtrl) GetRepoPoolManagers() map[string]common.PoolManager {
	return p.repositories
}
```
This is only a `get` function for a private field.

runner/repositories.go
mihaelabalutoiu: I will do the error checking.

runner/repositories.go
mihaelabalutoiu: I will check error here and wrap it further, if needed.

runner/organizations.go
mihaelabalutoiu: Same comment as the repo function.

runner/organizations.go
mihaelabalutoiu: Same comment as the repo function.

runner/runner.go
mihaelabalutoiu: I commented above why this is guaranteed to never fail. Let me know if that sounds good to you.

runner/interfaces.go
gabriel-samfira: If this can have a different implementation at some point in the future, that needs to query the database of some external source, we'll have to change it then to add an error. It's ok to return ```nil``` as a value in the current implementation. This is an interface definition that can have multiple implementations. We need to be mindful of future needs so we don't add technical debt needlessly.

runner/runner.go
mihaelabalutoiu: Same comment as previously.

runner/runner.go
mihaelabalutoiu: `GetRepoPoolManager()` replaces `findRepoPoolManager()`. 
It is only the function signature that differs, which requires an extra step to get the repo from store.
The `r.store.GetRepository(r.ctx, owner, name)` was previously part of `findRepoPoolManager()`.

I could keep `findRepoPoolManager()` on the runner if you want. But I wanted to refactor away all the pool manager functions for the runner.

runner/organizations.go
mihaelabalutoiu: I agree that the code with internal config is a bit repetitive. 
I will try to clean this up.

runner/runner.go
mihaelabalutoiu: Sounds great. I will add again `findRepoPoolManager` and `findOrgPoolManager`.

PR #8, opened 2022-08-19
Add `organizations.go` unit tests and `PoolManagerController` mocks

mihaelabalutoiu:
- Add `organizations.go` unit tests
  - Also, update vendor with the `github.com/stretchr/testify/suite`.
- Add `PoolManagerController` mocks


runner/organizations_test.go
```diff
@@ -22,19 +22,70 @@ import (
...
+		ctx:             adminCtx,
+		store:           storeMock,
+		poolManagerCtrl: poolMgrCtrlMock,
+	}
+
+	org, err := runner.CreateOrganization(adminCtx, createOrgParams)
+	storeMock.AssertExpectations(t)
+	poolMgrMock.AssertExpectations(t)
+	poolMgrCtrlMock.AssertExpectations(t)
+	require.Equal(t, createOrg, org)
```
gabriel-samfira, line 78: Comparing structs is a bit tricky. Especially if you have fields that point to other structs, or a slice of any kind. Slices are particularly tricky, because ordering matters when comparing. I am not sure if ```require.Equal()``` is enough in that case. For this canned result, I am certain you get the desired result, but it doesn't necessarily test the function properly. Estecially considering that the ```Organization{}``` structure also has a ```[]Pool``` field. It does test that we call the proper functions with expected args, which is good.

runner/organizations_test.go
```diff
@@ -22,19 +22,70 @@ import (
...
+	runnerMocks "garm/runner/mocks"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 )
 
+func TestCreateOrganization(t *testing.T) {
+	adminCtx := auth.GetAdminContext()
+	createOrgParams := params.CreateOrgParams{
```
gabriel-samfira, line 37: Setting up the dummy credentials and mocks feels a bit repetitive. Can we create a ```SetupTest()``` function or something similar?

```testify``` has the option to define a ```suite``` which gives you a ```SetupTest()``` function. This also allows you to create your own ```type``` and populate it on ```SetupTest()``` with any fixtures you may need, like credentials, responses, whatever makes sense.

You can use this to de-duplicate code.

https://github.com/stretchr/testify#suite-package

**Review by gabriel-samfira — CHANGES_REQUESTED** [review]
Before doing an in-depth review, it would help to deduplicate some code. Left a couple of comments that should be addressed first.

The rest of the changes are similar to the one I left the comments on, so reviewing those changes will be faster once the two comments are addressed.

Thanks for the PR!

File: runner/organizations_test.go (hunk 2 above)
Comment (mihaelabalutoiu) [reply runner/organizations_test.go:37]: Sounds good. I will look to remove the de-duplicate code.

File: runner/organizations_test.go (hunk 1 above)
Comment (mihaelabalutoiu) [reply runner/organizations_test.go:78]: You are right. I will update this as well.

Comment (mihaelabalutoiu) [reply runner/organizations_test.go]: Can you clarify this further ? Should I avoid `require.Equal` for structs ? If so, what should I use to compare structs against expected structs ?

From other tests that I've written, I noticed that `require.Equal` works fine comparing structs (even if they contain slices, other structs, etc).

File: runner/organizations_test.go [hunk 3]
```diff
@@ -22,170 +22,747 @@ import (
...
+
+	org, err := s.Runner.CreateOrganization(s.Fixtures.AdminContext, s.Fixtures.CreateOrgParams)
+
+	s.Fixtures.StoreMock.AssertExpectations(s.T())
+	require.Equal(s.T(), params.Organization{}, org)
+	require.Equal(s.T(), fmt.Sprintf("fetching org: %s", s.Fixtures.ErrMock.Error()), err.Error())
+}
+
+func (s *OrgTestSuite) TestCreateOrganizationAlreadyExists() {
+	s.Fixtures.CreateOrgParams.Name = "not-existent-name"
```
gabriel-samfira, line 173: Why change the name here, if we'll be returning the fixture anyway? We're setting up the mock to call ```GetOrganization``` with ```not-existent-name``` and returning an org from the fixture, which does not include a name. If you skip this, you should get the same result.

I am wondering if we should have a proper testing implementation of the database ```common.Store{}``` interface that actually creates those entities in memory, as normal store would, and return proper objects with all fields set. Or maybe just use the ```sql``` implementation with the ```sqlite``` backend, given the self contained nature of ```sqlite```.

The issue with mocking everything (duck typing) is that you run the risk of returning expectations that are not correct. Mocks may be outdated, or expectations may change. You may miss errors. It requires a lot of discipline to ensure mocks are correct or we risk tests lying to us. I am not suggesting that we never use mocks, but we should do so sparingly. Only use them where we have to, like when we need to call into the GitHub API or into the pool manager interface (which is somewhat opaque to this package).

runner/mocks/poolManagerController.go
```diff
@@ -0,0 +1,200 @@
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
gabriel-samfira, line 1: This file can be automatically generated using a ```go generate``` stanza. You can add it in ```runner/interfaces.go``` after the import statement like so:


```go
//go:generate mockery --name=PoolManagerController
```

You can then run:

```bash
go generate
```

in that folder or:

```bash
go generate ./...
```

in the root folder of the project, and it will automatically generate/update the mock.

gabriel-samfira (changes requested):
Looks a lot better. I know this may be a pain, but I think it's worth removing the ```store``` mock and simply using the ```sql``` database backend with the ```sqlite``` driver. It would spare you from a lot of mocks, and would yield better data you can test against (ie: objects that include all fields, including IDs)

runner/organizations_test.go
gabriel-samfira: For the test fixtures you use here, I am sure that comparing slices works just fine. However, it may fail when comparing unsorted slices. For example:

```go
require.Equal(t, [1,2,3], [3,2,1])
```

Even though the values are the same, the order differs. For cases where order matters, a failure is desirable. For cases where order does not matter, the error is a false positive. Also, this will most likely fail if one of the elements of a struct is a pointer. I am not sure if the ```require.Equal()``` walks through the struct and de-references pointers before comparing the value they point to. This is why we usually don't compare entire structs.

And in most cases, comparing the whole struct is not necessary. It is acceptable to compare values of individual fields we care about. In some cases, we only care that the value is non empty, as is the case of the auto-generated ```ID``` field in most objects returned by the API. We do care about fields such as ```Name``` in the case of an organization or repo. If we try to create an org and we request that it is called ```MyOrg``` then we expect the return value to have that name.

mihaelabalutoiu:
> Looks a lot better. I know this may be a pain, but I think it's worth removing the `store` mock and simply using the `sql` database backend with the `sqlite` driver. It would spare you from a lot of mocks, and would yield better data you can test against (ie: objects that include all fields, including IDs)

I made this changes. Please take a look now.

runner/organizations_test.go
```diff
@@ -19,173 +19,544 @@ import (
...
+	s.Fixtures.PoolMgrMock.AssertExpectations(s.T())
+	s.Fixtures.PoolMgrCtrlMock.AssertExpectations(s.T())
+	s.Require().Equal(fmt.Sprintf("starting org pool manager: %s", s.Fixtures.ErrMock.Error()), err.Error())
+}
+
+func (s *OrgTestSuite) TestListOrganizations() {
+	orgs, err := s.Runner.ListOrganizations(s.Fixtures.AdminContext)
+
+	s.Require().Nil(err)
+	s.Require().Equal(len(orgs), len(s.Fixtures.StoreOrgs))
```
gabriel-samfira, line 248: It may be useful to test that the expected names of the orgs match. Especially if there are only a couple of them. If this check is used in multiple tests, you can write a function to assert they match.

But this is not strictly necessary at this point.

runner/organizations_test.go
```diff
@@ -19,173 +19,544 @@ import (
...
+}
+
+func (s *OrgTestSuite) TestDeleteOrganization() {
+	s.Fixtures.PoolMgrCtrlMock.On("DeleteOrgPoolManager", mock.AnythingOfType("params.Organization")).Return(nil)
+
+	err := s.Runner.DeleteOrganization(s.Fixtures.AdminContext, s.Fixtures.StoreOrgs["test-org-3"].ID)
+
+	s.Fixtures.PoolMgrCtrlMock.AssertExpectations(s.T())
+	s.Require().Nil(err)
+	orgs, err := s.Fixtures.Store.ListOrganizations(s.Fixtures.AdminContext)
```
gabriel-samfira, line 277: A better way to test this is to call ```GetOrgByID()``` and make sure it returns ```ErrNotFound```

gabriel-samfira (changes requested):
Looks much better. Just one small nit and we can merge this.

mihaelabalutoiu:
> Looks much better. Just one small nit and we can merge this.

I made this change, please take a look now.

PR #11, opened 2022-09-08
Add more `repositories.go` unit testing

mihaelabalutoiu:
Add more `repositories.go` unit testing

### DIFF
--- a/runner/repositories_test.go
+++ b/runner/repositories_test.go
@@ -26,6 +26,7 @@ import (
 	"garm/runner/common"
 	runnerCommonMocks "garm/runner/common/mocks"
 	runnerMocks "garm/runner/mocks"
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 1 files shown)


runner/repositories_test.go
```diff
@@ -51,6 +54,22 @@ type RepoTestSuite struct {
 	Runner   *Runner
 }
 
+func (s *RepoTestSuite) EqualReposByName(expected map[string]params.Repository, actual []params.Repository) {
```
gabriel-samfira, line 57: This doesn't need to be exported if you don't want to call this function from another package. Functions that start with a capital letter are exported. If you'll just use it in this package it can be renamed to ```equalReposByName()```

gabriel-samfira: It's easier to compare two sorted slices instead of a map and a slice. To sort a slice you can use:

https://pkg.go.dev/sort#Slice

gabriel-samfira (changes requested):

runner/repositories_test.go
mihaelabalutoiu: Thank you for the suggestion. I implemented this and I updated the PR.

runner/repositories_test.go
```diff
@@ -51,6 +55,25 @@ type RepoTestSuite struct {
...
+	return reposSlice
+}
+
+func (s *RepoTestSuite) equalReposByName(expected, actual []params.Repository) {
+	s.Require().Equal(len(expected), len(actual))
+
+	sort.Slice(expected, func(i, j int) bool { return expected[i].Name > expected[j].Name })
+	sort.Slice(actual, func(i, j int) bool { return actual[i].Name > actual[j].Name })
+
+	for i := 0; i < len(expected); i++ {
```
gabriel-samfira, line 72: You also need to test that the lengths of both slices match. The ```expected``` array may be shorter, so you stop checking after you run out of elements. But in reality they are not equal.

gabriel-samfira: scratch that. You do check above. My mistake!

PR #16, opened 2022-09-15
Add `instances.go` unit tests

mihaelabalutoiu:
Add `instances.go` unit tests

### DIFF
--- a/database/sql/instances_test.go
+++ b/database/sql/instances_test.go
@@ -1,0 +1,171 @@
+// Copyright 2022 Cloudbase Solutions SRL
+//
+//    Licensed under the Apache License, Version 2.0 (the "License"); you may
+//    not use this file except in compliance with the License. You may obtain
+//    a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


database/sql/instances_test.go
```diff
@@ -0,0 +1,155 @@
...
+	// call tested function
+	instance, err := s.Store.CreateInstance(context.Background(), s.Fixtures.Pool.ID, createPoolParams)
+
+	// assertions
+	s.Require().Nil(err)
+	storeInstance, err := s.Store.GetInstanceByName(context.Background(), instanceName)
+	if err != nil {
+		s.FailNow(fmt.Sprintf("failed to get instance: %v", err))
+	}
+	s.Require().Equal(storeInstance.Name, instance.Name)
```
gabriel-samfira, line 114: we should test some more fields like pool ID, provider ID. Some of the fields we explicitly used when creating the entity in the database. Things of this sort. The idea is to test we get back a valid entity after creating it.

gabriel-samfira (changes requested):

PR #18, opened 2022-09-15
Cleanup unused struct field and Fix typo

mihaelabalutoiu:
* Cleanup unused struct field
  * Remove `Pool` field from `CreateInstanceParams` struct, because
this is given as a separate parameter to the `CreateInstance` function.
* Fix typo


mihaelabalutoiu:
I also tested the application with this update, and everything works fine.

PR #22, opened 2022-09-21
Add more `instances.go` unit tests

mihaelabalutoiu:
Add more `instances.go` unit tests

Signed-off-by: Mihaela Balutoiu <mbalutoiu@cloudbasesolutions.com>

### DIFF
--- a/database/sql/instances_test.go
+++ b/database/sql/instances_test.go
@@ -20,6 +20,8 @@ import (
 	dbCommon "garm/database/common"
 	garmTesting "garm/internal/testing"
 	"garm/params"
+	"garm/runner/providers/common"
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
@@ -35,6 +37,17 @@ type InstancesTestSuite struct {
 	suite.Suite
 	Store    dbCommon.Store
 	Fixtures *InstancesTestFixtures
+}
+
+func (s *InstancesTestSuite) equalInstancesByName(expected, actual []params.Instance) {
+	s.Require().Equal(len(expected), len(actual))
+
+	sort.Slice(expected, func(i, j int) bool { return expected[i].Name > expected[j].Name })
+	sort.Slice(actual, func(i, j int) bool { return actual[i].Name > actual[j].Name })
+
+	for i := 0; i < len(expected); i++ {
+		s.Require().Equal(expected[i].Name, actual[i].Name)
+	}
 }
 
 func (s *InstancesTestSuite) SetupTest() {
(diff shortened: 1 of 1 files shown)


database/sql/instances_test.go
```diff
@@ -73,10 +86,12 @@ func (s *InstancesTestSuite) SetupTest() {
 			context.Background(),
 			pool.ID,
 			params.CreateInstanceParams{
-				Name:        fmt.Sprintf("test-instance-%v", i),
-				OSType:      "linux",
-				OSArch:      "amd64",
-				CallbackURL: "https://garm.example.com/",
+				Name:         fmt.Sprintf("test-instance-%v", i),
```
gabriel-samfira, line 89: I think you can replace ```%v``` with ```%d```

database/sql/instances_test.go
```diff
@@ -164,6 +179,115 @@ func (s *InstancesTestSuite) TestGetInstanceByNameFetchInstanceFailed() {
...
+	storeInstance := s.Fixtures.Instances[0]
+	err := s.Store.DeleteInstance(context.Background(), s.Fixtures.Pool.ID, storeInstance.Name)
+
+	s.Require().Nil(err)
+
+	_, err = s.Store.GetPoolInstanceByName(context.Background(), s.Fixtures.Pool.ID, s.Fixtures.Instances[0].Name)
+	s.Require().Equal("fetching instance: fetching pool instance by name: not found", err.Error())
+}
+
+func (s *InstancesTestSuite) TestDeleteInstanceFailed() {
```
gabriel-samfira, line 192: This mostly tests if ```DeleteInstance()``` validates the pool ID. Should be renamed to something like: ```TestDeleteInstanceInvalidPoolID```

Same with a few more functions bellow.

gabriel-samfira: If you want to test a condition where the database returns an error, you have two options:

  * Use mocks
  * Make/open the database read-only and try to mutate the data

gabriel-samfira (changes requested):

database/sql/instances_test.go
mihaelabalutoiu: > This mostly tests if `DeleteInstance()` validates the pool ID. Should be renamed to something like: `TestDeleteInstanceInvalidPoolID`
> 
> Same with a few more functions bellow.

Fixed

mihaelabalutoiu: > If you want to test a condition where the database returns an error, you have two options:
> 
> * Use mocks
> * Make/open the database read-only and try to mutate the data

Sounds good. I will add such a test case in a separate PR. Thank-you for the suggestion.

Issue #25, opened 2022-10-05
Webhook owner match should be case insensitiv.


maigl:
# The issue

Webhook requests return with `404` when the `org` or the `repo name` do not have the exact same casing as defined my garm runner pool. In my concret case my pool is registered for an organization "org" but in github this org is spelled "OrG". 

Github supports mixed spelled `org` and `repo` names. But all github apis work case-insensitive. 
E.g. see:
https://docs.github.com/en/rest/orgs/orgs 
```
Path parameters
--

orgstringRequiredThe organization name. The name is not case sensitive.

```

# Proposed solution

- [ ] All owner validation should be performed in lower case.
- [ ] All reads and writes from and to the store should be converted to lower case.

# Remarks

I could send a pull request if you like.

<sub>Michael Kuhnt <michael.kuhnt@mercedes-benz.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))
</sub>

PR #26, opened 2022-10-05
Make owners and repos case insensitive

gabriel-samfira:
Github treats owners (users and orgs) and repos as case insensitive. To github, https://github.com/cloudbase/garm is equivalent to https://github.com/CloudBase/GaRm. This commit makes the sql store backend, case insensitive when querying repos and orgs.

Fixes: #25


maigl:
Hi @gabriel-samfira, very nice .. but I think we also need to make the comparison in ValidateOwner case insensitive. The reason is that the owner in the workflow_job event will have some case and we should make it robust here.

maigl (approved):
lgtm :+1:

gabriel-samfira:
Tested with both orgs and repos:

```bash
ubuntu@garm:~$ garm-cli repo ls
+--------------------------------------+-----------------+---------+------------------+
| ID                                   | OWNER           | NAME    | CREDENTIALS NAME |
+--------------------------------------+-----------------+---------+------------------+
| f0b1c1c8-b605-4560-adb7-79b95e2e470c | gABrIeL-SaMfIra | ScRiPtS | gabriel          |
+--------------------------------------+-----------------+---------+------------------+
ubuntu@garm:~$ garm-cli org ls
+--------------------------------------+----------+------------------+
| ID                                   | NAME     | CREDENTIALS NAME |
+--------------------------------------+----------+------------------+
| 3f7b0a5c-1d7a-4e52-b0c8-06116eca4091 | GsAmFiRa | gabriel_org      |
+--------------------------------------+----------+------------------+
```

Seems to work well. Merging.

@maigl let me know if it works for you.

PR #49, opened 2022-12-05
Add a metadata resource for instances

gabriel-samfira:
Currently, garm leverages the user data feature of providers to create runners and set them up. Storing the runner registration token may be a security vulnerability in some situations (see https://github.com/cloudbase/garm/discussions/48). This PR adds a new API endpoint that can be used by instances to retrieve a github runner registration token.

This new endpoint uses the same authentication middleware as the status update endpoint used by the instances to call home and report their state. The authentication middleware has also been changed to only allow access to instances that are in the ```pending``` or ```installing``` state. once the instance transitions to any other state (```installed```, ```failed```, etc), the instance can no longer access the two resources.


gabriel-samfira:
An additional safeguard was added to limit the number of tokens that can be requested by a bearer token belonging to a particular instance. If an instance token has fetched a github runner registration token, any subsequent attempt to fetch another runner token will fail.

PR #60, opened 2023-01-19
feat: add prometheus metrics & endpoint

maigl:
This adds a prometheus metrics endpoint and 2 simple metrics to allow garm being monitored.

- The first metrics shows which runner instances currently exist and their status(es).
- The second metric counts the webhooks garm receives.

Remarks:
- the funcs `ListAllInstances` and `ListAllPools` now have a version without the admin check that can be used internally.
  I didn't want to fake admin 

</sub>


gabriel-samfira:
This looks amazing!

I need to get some proper sleep before looking at this. I will definitely do a full review tomorrow morning!

For now, just one note in regards to the need for ```ListAllInstances()``` and ```ListAllPools()``` for internal use. We have the ```auth.GetAdminContext()``` function that was made specifically for this purpose. It is intended to allow us to use these functions internally without needing to wrap them. It's available here:

https://github.com/cloudbase/garm/blob/main/auth/context.go#L219-L227

It's mainly used for testing now, but the intention for it was to be used for your exact use case.

apiserver/routers/routers.go
```diff
@@ -23,13 +23,20 @@ import (
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 func NewAPIRouter(han *controllers.APIController, logWriter io.Writer, authMiddleware, initMiddleware, instanceMiddleware auth.Middleware) *mux.Router {
 	router := mux.NewRouter()
 	logMiddleware := util.NewLoggingMiddleware(logWriter)
 	router.Use(logMiddleware)
 
+	// Handles github webhooks
+	router.Handle("/metrics", promhttp.Handler())
```
gabriel-samfira, line 37: I think we should create a sub router for the metrics endpoint. We could then apply any kind of middleware we want just on that router. 

Something like:

```suggestion
	metricsRouter := router.PathPrefix("/metrics").Subrouter()
	metricsRouter.Handle("/", promhttp.Handler())
```

Which brings me to the authentication side of things. Can we add any kind of authentication? Even basic auth would do. Prometheus seems to allow the ```Authorization``` header to be set in the ```scrape_config``` section, which means we can use the regular ```JWT``` token auth. Or at least basic auth.

Not very comfortable exposing metrics publicly.

gabriel-samfira (commented):
Added a couple of comments. Will look at everything a bit closer tomorrow. This is a great feature to have!

apiserver/routers/routers.go
gabriel-samfira: I was thinking of creating a new auth middleware for metrics, limited in scope, just like the instance callback middleware. We could add an API endpoint where we could generate tokens for the metrics service, and you could do something like:

```bash
garm-cli metrics token create
```

With optional parameters to limit the validity time, and possibly other context fields, which you could then configure in your Prometheus ```scrape_config```.

These tokens would only work against ```/metrics```. The ```/metrics``` endpoint would be accessible to anyone with an ```admin``` context or the limited ```metrics``` token. This way we could possibly (later) expose the metrics to the garm CLI as well.

I can have a stab at this if you'd like. Would mandatory auth be an issue for your setup?

gabriel-samfira: As a general thing, I tend to treat other systems as adversarial. Even if they are in the same network and even if there are some firewall rules in place to limit which systems can talk to other systems. If there is any chance of exposing any kind of sensitive information, I usually do it after checking that the remote system was granted explicit access to do so.

maigl: that is fine for me .. but probably you'll have to take a lead in implementing this.

Still, I think it should be possible to disable metrics auth. Normally, e.g. in a k8s cluster, some (probably managebd by somebody else) prometheus automatically discovers 'things' to scrape. And this would be impossible if it needs to authenticate itself.

apiserver/controllers/controllers.go
```diff
@@ -110,14 +145,23 @@ func (a *APIController) handleWorkflowJobEvent(w http.ResponseWriter, r *http.Re
...
 	hookType := r.Header.Get("X-Github-Hook-Installation-Target-Type")
 
+	hostname, controllerId := a.GetControllerInfo()
+
 	if err := a.r.DispatchWorkflowJob(hookType, signature, body); err != nil {
 		if errors.Is(err, gErrors.ErrNotFound) {
+			webhooksReceived.WithLabelValues("false", "owner_unknown", hostname, controllerId).Inc()
 			log.Printf("got not found error from DispatchWorkflowJob. webhook not meant for us?: %q", err)
 			return
+		} else if strings.Contains(err.Error(), "signature") {
```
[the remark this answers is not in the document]
maigl: btw .. this error never come. There is a bug in the webhook validation which has the effect that every webhook for orgs end enterprises are always valid. see #61

apiserver/routers/routers.go
gabriel-samfira: We can have it without auth to start with, but I think we should document this [in the docs folder](/cloudbase/garm/tree/main/doc), and also amend the sample ```nginx``` config file to set up a separate ```location /metrics``` as an example of how people could put this behind a reverse proxy and potentially add basic auth using nginx.

What do you think?

gabriel-samfira: Apropos, the [prometheus scape_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cscrape_config%3E) has the [tls_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tls_config) section which allows you to set the client certificates.

There is an example here:

https://github.com/prometheus/prometheus/blob/main/documentation/examples/prometheus-kubernetes.yml

apiserver/controllers/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+				Type: "repository",
+			}
+		}
+	}
+
+	hostname, controllerID := c.apiController.GetControllerInfo()
+
+	for _, instance := range instances {
+
+		ch <- prometheus.MustNewConstMetric(
```
gabriel-samfira, line 80: Does this have the potential to panic?

By convention, functions that contain ```Must``` usually panic if something is wrong. While this may be ok in ```init()``` functions, because you see that something is terribly wrong when you run the app, calling a function that can panic is not a good idea elsewhere. We would need to ```recover()``` somehow, or risk crashing garm.

runner/pool/pool.go
```diff
@@ -401,7 +398,7 @@ func (r *basePoolManager) AddRunner(ctx context.Context, poolID string) error {
 	if prefix == "" {
 		prefix = params.DefaultRunnerPrefix
 	}
-	name := fmt.Sprintf("%s-%s", prefix, shortid.MustNew(0, shortidABC, 42).String())
+	name := fmt.Sprintf("%s-%s", prefix, shortid.MustGenerate())
```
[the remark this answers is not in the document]
maigl: .. which directly happend! I'll change that!

apiserver/controllers/metrics.go
gabriel-samfira: There is an equivalent without the ```Must``` bit. You can log the error and return/continue. But it must not panic.

apiserver/controllers/controllers.go
```diff
@@ -110,14 +145,23 @@ func (a *APIController) handleWorkflowJobEvent(w http.ResponseWriter, r *http.Re
 	signature := r.Header.Get("X-Hub-Signature-256")
 	hookType := r.Header.Get("X-Github-Hook-Installation-Target-Type")
 
+	hostname, controllerId := a.GetControllerInfo()
```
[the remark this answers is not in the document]
maigl: in case of `hostname` I disagree a little bit. The hostname can change at any time, it's not in our hands. I would be nice if the metrics would reflect this.

apiserver/routers/routers.go
maigl: .. it's possible in prometheus, but if this prometheus is managed by a 3rd party and scans (a potentially big) number of services this 3rd party would have to handle (store, delete, update) many customer secrets. 
I think in reality in such a setup prometheus scrapes only endpoints that have no own auth but are not exposed anywhere. prometheus then uses kubeproxy to discover and scrape them.

Shall we start exposing metrics on an own port? and try to offer optional auth for that in a 2nd step ?

apiserver/controllers/controllers.go
```diff
@@ -82,6 +88,32 @@ func handleError(w http.ResponseWriter, err error) {
...
+	// the hostname is neither fixed nor in our control
+	// so we get it every time to avoid confusion
+	var err error
+	hostname, err = os.Hostname()
+	if err != nil {
+		log.Printf("error getting hostname: %q", err)
+		return "", ""
+	}
+
+	if a.id == "" {
```
gabriel-samfira, line 104: I think you can do this in ```NewAPIController()```, and skip this check here. You can error our in ```NewAPIController()``` if you can't get the ID. There is a call to ```maybeInitController()``` in ```main.go``` which ensures that by the time we call ```NewAPIController()``` there is an ID generated.

apiserver/routers/routers.go
gabriel-samfira: I think we can get away with putting it behind a reverse proxy like nginx for now. In nginx we can define a ```location /metrics``` that proxy passes back to garm, and in that ```location``` block if anyone wants to, can enable basic auth or whatever other kind of restrictions.

I can have a look at creating a middleware for the ```/metrics``` endpoint next week.

maigl:
I started to implement this auth feature with jwt for metrics .. so converting this PR to draft until I'm done .. should not take to long :)

apiserver/controllers/metrics.go
maigl: done

maigl:
Hi,

I changed the code so now you can:
- set a port for metrics, must be different from main port
- disable metrics endpoint 
- disable auth for metrics endpoint

the cli now has
```
garm-cli metrics-token create
```
with this git you can curl the metrics endpoint, but not the normal server endpoints, vice versa the admin token `cannot` curl the metrics endpoint

I skipped adding a dedicated metrics user in the db.

the configuration in the garm/config.toml looks like this:
```
[apiserver]
  bind = "0.0.0.0"
  port = 9997
  use_tls = false
  cors_origins = ["*"]
  [apiserver.metrics]
    no_auth = false
    port = 8082
    disabled = false
....

```

gabriel-samfira:
I think we can skip having to add another listener for metrics. In some cases, that's just another port to justify to the security team, and while it does seem to be optional, I think it's best we don't add a moving part unless absolutely unavoidable.

In this case, if we really want to, we can do this by putting nginx in front of garm. Garm can listen on ```127.0.0.1``` and a reverse proxy can be configured in whichever fashion the user wants. They can create a new ```server{}``` block on a different port just for ```location /metrics```  or they can restrict which remote IP can access that ```location```. If you'd like, I can create a PR against your branch with a sample.

Our  job should be to make sure that authentication and authorization is properly fleshed out. Tokens need to have just enough privileges to do the things they need to do. There's no need for a dedicated "metrics" user. In fact, the user ID can be left blank, or a new claims type can be created. At a later time, it may be useful to create a unified claims type that allows us to set access levels. At this point, we can simply add a new field in the claims that states the token is only valid for metrics. We parse the token, and set the context appropriately.

Almost all current handlers and runner functions check if the context is that of ```admin```. There are also a few functions that allow instances to call home and set their own status.

Same can be done for metrics.

I will look over this PR today in more detail and can give more constructive feedback point by point. But this is shaping up nicely and is a great addition to garm. Can't wait to see it in action :smile:.

p1
```diff
@@ -0,0 +1,2 @@
+Identity added: /home/mikuhnt/.ssh/id_rsa (/home/mikuhnt/.ssh/id_rsa)
```
gabriel-samfira, line 1: I am guessing this should not be here :smile:

maigl: damn .. thx

gabriel-samfira:
added some nginx samples: https://github.com/mercedes-benz/garm/pull/3

apiserver/controllers/controllers.go
```diff
@@ -180,6 +232,19 @@ func (a *APIController) NotFoundHandler(w http.ResponseWriter, r *http.Request)
 	json.NewEncoder(w).Encode(apiErr)
 }
 
+func (a *APIController) MetricsTokenHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
```
gabriel-samfira, line 237: We should check ```if auth.IsAdmin(ctx)```. Both here and in ```GetJWTMetricsToken()```, even if we do check that in the auth middleware. We need to guard against someone using this in a different code path and forgets to properly validate before calling these functions. Call me paranoid, but better safe than sorry.

apiserver/controllers/controllers.go
```diff
@@ -180,6 +232,19 @@ func (a *APIController) NotFoundHandler(w http.ResponseWriter, r *http.Request)
...
+	token, err := a.auth.GetJWTMetricsToken(ctx)
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintf(w, `{"token": "%s"}`, token)
```
gabriel-samfira, line 245: There is a ```JWTResponse{}``` in params. Please use that.

```suggestion
	json.NewEncoder(w).Encode(runnerParams.JWTResponse{Token: token})
```

cmd/garm/main.go
```diff
@@ -166,6 +168,28 @@ func main() {
 		}
 	}()
 
+	if !cfg.APIServer.MetricsConfig.Disabled {
+		go func() {
+
+			metricsMiddleware := auth.NewMetricsMiddleware(cfg.JWTAuth)
+
+			r := mux.NewRouter()
```
gabriel-samfira, line 176: Please use the router inside the ```apiserver/router``` package. You can create a new sub-router from the main router. That router has only the logging middleware attached.

cmd/garm/main.go
```diff
@@ -166,6 +168,28 @@ func main() {
...
+			if !cfg.APIServer.MetricsConfig.NoAuth {
+				r.Use(metricsMiddleware.Middleware)
+			}
+
+			err := prometheus.Register(controllers.NewGarmCollector(controller))
+			if err != nil {
+				log.Printf("failed to register prometheus collector: %+v", err)
+			}
+
+			if err := http.ListenAndServe(cfg.APIServer.MetricsBindAddress(), r); err != nil {
```
gabriel-samfira, line 187: Please use the existing listener. If a separate port is desired for metrics, nginx can be used as a reverse proxy to tweak how garm is exposed.

config/config.go
```diff
@@ -476,14 +476,30 @@ func (t *TLSConfig) Validate() error {
 	return nil
 }
 
+type MetricsConfig struct {
```
gabriel-samfira, line 479: Please add a ```Validate()``` function to this config type. All config types have validation.

config/config.go
```diff
@@ -476,14 +476,30 @@ func (t *TLSConfig) Validate() error {
...
-	Port        int       `toml:"port" json:"port"`
-	UseTLS      bool      `toml:"use_tls" json:"use-tls"`
-	TLSConfig   TLSConfig `toml:"tls" json:"tls"`
-	CORSOrigins []string  `toml:"cors_origins" json:"cors-origins"`
+	Bind          string        `toml:"bind" json:"bind"`
+	Port          int           `toml:"port" json:"port"`
+	UseTLS        bool          `toml:"use_tls" json:"use-tls"`
+	TLSConfig     TLSConfig     `toml:"tls" json:"tls"`
+	CORSOrigins   []string      `toml:"cors_origins" json:"cors-origins"`
+	MetricsConfig MetricsConfig `toml:"metrics" json:"metrics"`
```
gabriel-samfira, line 502: The metrics config deserves a place in the ```Config{}``` struct. Or in the ```Default{}``` struct.

runner/runner.go
```diff
@@ -265,6 +266,14 @@ type Runner struct {
 	credentials map[string]config.Github
 }
 
+func (r *Runner) GetControllerID() (uuid.UUID, error) {
```
gabriel-samfira, line 269: This can be removed if you give the collector access to the store.

As a side note, all functions should accept ```context.Context```, for at least the following reasons:

  * The context may get canceled and if we pass it along, it allows the various libraries we use to cancel their tasks and return gracefully.
  * It allows us to do some authorization checks

apiserver/controllers/controllers.go
```diff
@@ -29,9 +32,15 @@ import (
 
 	"github.com/gorilla/websocket"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 func NewAPIController(r *runner.Runner, auth *auth.Authenticator, hub *wsWriter.Hub) (*APIController, error) {
+	id, err := r.GetControllerID()
```
gabriel-samfira, line 39: The controller ID could be passed as an argument to ```NewAPIController()``` from ```main()```. It spares us from calling this function, as we already have that info in ```main()```.

maigl:
Thank you, I'm a bit busy next week, but I'll fix everything as soon as I can.

gabriel-samfira:
> Thank you, I'm a bit busy next week, but I'll fix everything as soon as I can.

No worries at all! Thanks for all the work! This will be a great addition to garm!

gabriel-samfira:
I fixed some linting errors and added a new workflow that checks things like gofmt, linting, vendoring consistency, etc. There are now a few makefile targets that should help out.

```bash
make verify
```

will run the lint, gofmt, vendor checks.

```bash
make test
```

will run verify and the go tests.

This is the last time I cause conflicts for your PR :smile:. Sorry about that!

apiserver/controllers/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		} else {
+			poolNames[pool.ID] = poolInfo{
+				Name: pool.RepoName,
+				Type: "repository",
+			}
+		}
+	}
+
+	hostname, controllerID := c.apiController.GetControllerInfo()
```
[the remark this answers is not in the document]
maigl: the func `apicontroller.GetControllerInfo()` should be a nice helper, as I probably need `hostname` and `controller` id in all metrics (also future ones). I thought it would be clumsy to gather those values in all particulare places where I want to increment some metrics counter.
Example handleWorkflowJobEvent: this func needs hostname and controller id only for annotating the metrics, so I wanted to add as little as possible code to get those details.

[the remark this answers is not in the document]
maigl: I removed the dependency in to apicontroller in the collector .. but still I think this func makes sense

apiserver/controllers/controllers.go
maigl: If I see it correctly we don't have the `id` directly in main, what am I missing ?

apiserver/controllers/controllers.go
```diff
@@ -82,6 +94,22 @@ func handleError(w http.ResponseWriter, err error) {
 	json.NewEncoder(w).Encode(apiErr)
 }
 
+// GetControllerInfo returns means to identify this very garm instance.
+// This is very useful for debugging and monitoring purposes.
+func (a *APIController) GetControllerInfo() (hostname, controllerId string) {
```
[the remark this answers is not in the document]
maigl: this func is supposed to be a nice helper, as I probably need hostname and controller id in all metrics (also future ones). I thought it would be clumsy to gather those values in all particulare places where I want to increment some metrics counter.
Example handleWorkflowJobEvent: this func needs hostname and controller id only for annotating the metrics, so I wanted to add as little as possible code to get those details, I hope this could help keep good readability?!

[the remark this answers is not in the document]
gabriel-samfira: Then how about this:

  * We extend: https://github.com/cloudbase/garm/blob/a0670c650352037bdabd5a53a0371fa21620d8b2/params/params.go#L276-L278 to include the controller hostname.
  * We create a function in ```runner.Runner{}``` that exposes the controller info, with the usual context validation, making sure whomever retrieves that info has access to do so. If we need that info internally, we can always use an admin context.
    * Checking authorization in all ```runner.Runner{}``` functions will help us in the future as well if we decide to split up components. In a zero trust environment (any remote system should be considered adversarial to be hones) we need to check if the remote **authenticated** component is authorized to perform a specific action. Right now, in most places we check for ```admin```, but with minimal planning, we can extend this to do much more complex validation in the future.
  * We populate the new hostname field with a call to ```os.Hostname()``` in ```runner.ControllerInfo(ctx context.Context)```
  * The ```runner.Runner{}``` type should have the ```controllerID``` cached. This way we don't need to hit the DB for a value that should never change. All we need is the hostname.
  * Use the ```runner.Runner{}``` we inject in the API controller (should already be there)/metrics collector to fetch that info.

This should keep concerns separated and readability at an acceptable level. 

(side note: If we really want to, we can define interfaces with a more narrow scope, that the ```runner.Runner{}``` type implements, and pass that along as a dependency to whichever components need it.)

I would like to avoid crossing boundaries.

  * The API controller should be used only as a presentation layer to the API consumer. It should not be used as a dependency in anything else.
  * The ```runner.Runner{}``` should be used as a central controller the API, the pool manager and any other component that needs to fetch info from the store or fetch jobs it needs to perform, calls into. This **should** be the only layer with direct access to the store.
  * The pool manager is the layer that **should** call into the ```runner.Runner{}``` to get the jobs it needs to perform to make sure that what we need is applied to the providers.
  * The providers are the glue between garm and the IaaS.

I realize this is not the case at the moment, and that we *do* cross boundaries right now, but I would like to avoid doing that more than necessary. The less we have to refactor later, the better.

The plan is to split these components up at some point, which will allow us to scale better if we need to. At that point we can replace the current dependencies we inject with a compatible ```RPC``` implementation that exposes the same functionality. This will hopefully allow us to not break everything when we split things up.

So it would be really helpful to group logic as well as possible and not spread it across multiple boundaries. I've seen this happen in the past, and it leads to spaghetti nightmare.

Does this sound sensible? Any suggestions are welcome.

runner/runner.go
maigl: I didn't give the collector access directly to the store, for now this is not necessary, access to the runner seems to be a good level of abstraction.

apiserver/controllers/controllers.go
gabriel-samfira: We have this bit:

https://github.com/cloudbase/garm/blob/main/cmd/garm/main.go#L104-L106

which is defined here: https://github.com/cloudbase/garm/blob/main/cmd/garm/main.go#L51-L61

There is no reason why it could not return the controller info as well as the error. You could then pass that info along to whichever component needs it.

runner/runner.go
gabriel-samfira: I agree. See above wall of text :smile:

config/config.go
maigl: since you want to use the same router, there is no port to validate .. so basically the only real config toggle is a bool (no_auth). Validation makes not much sense here ..
do you still want a validate func ?

btw, I removed the `disable metrics` toggle, mainly because in the router I don't have access to the config and I don't know if one would want to disable metrics. What do you think? shall we be able to disable metrics? 
viper would really be nice here :)

gabriel-samfira: > viper would really be nice here :)

We'll get there. Eventually :smile:. It's not worth the effort for now.

> do you still want a validate func ?

It's a good idea to have a ```Validate()``` func, even if it's a noop, **if we have a separate type** for a particular set of settings. If we embed the metrics auth in an existing type, and it's a bool, there is no need to validate it.

The idea is that we may extend a config at some point, and may forget to add a ```Validate()``` func later on.

> btw, I removed the `disable metrics` toggle, mainly because in the router I don't have access to the config and I don't know if one would want to disable metrics.

You can pass in a new param to ```NewAPIRouter()``` which includes whichever config you need. It didn't make sense to pass in a config to that func until now, but if that changes, I have nothing against adding a param that makes sense. 

I think it may be useful to have a toggle to disable metrics.

cmd/garm/main.go
maigl: done .. but I had to pass the config into the apirouter to allow enabling the metrics and the need for auth

config/config.go
maigl: .. I now pass the config into NewAPIRouter

config/config.go
```diff
@@ -476,6 +477,11 @@ func (t *TLSConfig) Validate() error {
 	return nil
 }
 
+type Metrics struct {
+	NoAuth  bool `toml:"no_auth" json:"no-auth"`
```
gabriel-samfira, line 481: What do you think about the following defaults for metrics:

  * By default they are disabled
  * Authentication is enabled by default

In which case the settings can look something like this:

```suggestion
	DisableAuth  bool `toml:"disable_auth" json:"disable-auth"`
	Enabled        bool `toml:"enabled" json:"enabled"`
```

That way, if this section is completely omitted,  defaults should be good.

Considering this is just a couple of boolean values, you're right, makes no sense to validate.

gabriel-samfira: A config would look something like this:

```toml
[metrics]
enabled = true
disable_auth = false
```

apiserver/controllers/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		log.Printf("cannot collect metrics, getting hostname: %s", err)
+	}
+
+	c.CollectInstanceMetric(ch, hostname, controllerID.String())
+	c.CollectHealthMetric(ch, hostname, controllerID.String())
+}
+
+func (c *GarmCollector) CollectHealthMetric(ch chan<- prometheus.Metric, hostname string, controllerID string) {
+	ch <- prometheus.MustNewConstMetric(
```
gabriel-samfira, line 56: This might panic.

gabriel-samfira:
The last two comments are the last nitpicking I have :smile:

cmd/garm/main.go
```diff
@@ -137,7 +143,16 @@ func main() {
...
 	}
 
-	router := routers.NewAPIRouter(controller, multiWriter, jwtMiddleware, initMiddleware, instanceMiddleware)
+	metricsMiddleware, err := auth.NewMetricsMiddleware(cfg.JWTAuth)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = prometheus.Register(controllers.NewGarmCollector(runner))
+	if err != nil {
+		log.Printf("failed to register garm collector in prometheus", err)
```
gabriel-samfira, line 152: It's complaining about the missing : ```%s``` in the ```log.Printf()``` statement.

apiserver/controllers/controllers.go
```diff
@@ -170,6 +215,23 @@ func (a *APIController) NotFoundHandler(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	token, err := a.auth.GetJWTMetricsToken(ctx)
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(runnerParams.JWTResponse{Token: token})
```
gabriel-samfira, line 232: It's complaining that we're throwing away the error here. I put a ```log.Printf``` everywhere else, to see when such an error occurs, but feel free to add ```// nolint``` at the end of the line.

apiserver/controllers/metrics.go
maigl: .. sorry for the confusion .. I thought I hit 're-request review' once I'm done .. but you're too fast

gabriel-samfira: I will let you finish :smile:. Will check back later!

maigl:
.. so .. finally, I think you should be good to do another review ..

runner/runner.go
```diff
@@ -263,6 +264,27 @@ type Runner struct {
 
 	providers   map[string]common.Provider
 	credentials map[string]config.Github
+
+	controllerInfo params.ControllerInfo
+}
+
+func (r *Runner) GetControllerInfo(ctx context.Context) (params.ControllerInfo, error) {
+	if r.controllerInfo == (params.ControllerInfo{}) {
```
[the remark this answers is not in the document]
gabriel-samfira: Everything else looks fine!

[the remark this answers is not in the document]
gabriel-samfira: Ahh. My mistake. ```os.Hostname()``` can return an error. So that would be something like:

```go
hostname, err := os.Hostname()
if err != nil {
    return params.ControllerInfo{}, errors.Wrap(err, "getting hostname")
}
return params.ControllerInfo{
  ControllerID: r.poolManagerCtrl.controllerID,
  Hostname: hostname,
}, nil
```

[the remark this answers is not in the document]
maigl: hm .. accessing the id through Runner.poolManagerCtrl.controllerID is not that easy .. as you decoupled the access via an interface, I didn't want to extend this .. and type assertion seems not the right way .. 

os.Hostname will probably not really error .. and if so we will ignore it anyway .. we cannot crash, right? 
So I went for having no error result, but there's still the context in the runner.GetControllerInfo .. do you think that still makes  much sense ?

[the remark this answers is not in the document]
gabriel-samfira: You are right. Then it might be worth adding a new un-exported field to ```Runner{}``` like ```controllerID uuid.UUID```, and cache the controller ID there for later use. 

Note, the ```NewRunner()``` function is only called once. If you want to catch a hostname change, you probably want to call that in ```GetControllerInfo()```. The call to ```os.Hostname()``` shouldn't error, but if they added that value as a return, it *may* happen, in which case you risk getting an empty string there. Is that ok in your current code path?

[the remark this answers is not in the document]
gabriel-samfira: > So I went for having no error result, but there's still the context in the runner.GetControllerInfo .. do you think that still makes much sense ?

The context is useful if we want to limit who gets to fetch this info. There is no real sensitive info returned by this function, but we should probably still check for authorization here. In the future we may extend this function to return more metadata about the controller, especially if we will decouple the code and possibly deploy multiple instances of that particular component.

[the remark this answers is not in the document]
gabriel-samfira: I will not nitpick too much on this. I am okay with merging as is, and making changes later as needed. The only thing I think is worth changing is the location of the call to ```os.Hostname()```. It's worth moving it into ```GetControllerInfo()``` as that should resolve the hostname on every call to ```GetControllerInfo()```, as opposed to caching it for the duration of the lifetime of the current process.

You mentioned previously that you would rather pick up on hostname changes without restarting garm.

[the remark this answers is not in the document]
maigl: > You mentioned previously that you would rather pick up on hostname changes without restarting garm.

right .. thanks for your reviews and your patience with me :smile:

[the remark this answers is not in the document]
gabriel-samfira: Thank you for the work you put into this!

PR #61, opened 2023-01-19
Fix webhook signature validation

matoubidou:
Webhooks seemed never to be validated (see behaviour for empty secret in [runner.go#L544-547](https://github.com/cloudbase/garm/blob/main/runner/runner.go#L544-547)) as the secret of the enterprises and orgs were not properly set (see [enterprise.go#L73-L86](https://github.com/cloudbase/garm/blob/main/database/sql/enterprise.go#L73-L86), [organizations.go#L71-L84](https://github.com/cloudbase/garm/blob/main/database/sql/organizations.go#L71-L84)). For repos the feature seems to work as expected.

**Essentially we need to decide whether empty secrets can be set to disable the validation or not**. After we decided on that I would update the PR accordingly.

* so far I only fixed the code for enterprises after we decided whether we allow for empty secrets  I will fix the code for organizations, too. Of course I can align the code for enterprises, organizations and repos.
* there have been three failing tests  that I could not fix because they depend on side effects in the current implementation. Do we need these tests? Their implementation would depend on the decision above.

The issue was recognize during the work on #60.

<sub>Mathias Zeller <mathias.zeller@mercedes-benz.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))</sub>

### DIFF
--- a/database/sql/enterprise_test.go
+++ b/database/sql/enterprise_test.go
@@ -20,12 +20,12 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"testing"
 
 	dbCommon "garm/database/common"
 	runnerErrors "garm/errors"
 	garmTesting "garm/internal/testing"
 	"garm/params"
-	"testing"
 
 	"github.com/stretchr/testify/suite"
 	"gopkg.in/DATA-DOG/go-sqlmock.v1"
@@ -194,7 +194,7 @@ func (s *EnterpriseTestSuite) TestCreateEnterpriseInvalidDBPassphrase() {
 		s.Fixtures.CreateEnterpriseParams.WebhookSecret)
 
 	s.Require().NotNil(err)
-	s.Require().Equal("failed to encrypt string", err.Error())
+	s.Require().Equal("encoding secret: invalid passphrase length (expected length 32 characters)", err.Error())
 }
 
 func (s *EnterpriseTestSuite) TestCreateEnterpriseDBCreateErr() {
@@ -238,16 +238,16 @@ func (s *EnterpriseTestSuite) TestGetEnterpriseNotFound() {
 }
 
 func (s *EnterpriseTestSuite) TestGetEnterpriseDBDecryptingErr() {
-	s.Fixtures.SQLMock.
-		ExpectQuery(regexp.QuoteMeta("SELECT * FROM `enterprises` WHERE name = ? COLLATE NOCASE AND `enterprises`.`deleted_at` IS NULL ORDER BY `enterprises`.`id` LIMIT 1")).
-		WithArgs(s.Fixtures.Enterprises[0].Name).
-		WillReturnRows(sqlmock.NewRows([]string{"name"}).AddRow(s.Fixtures.Enterprises[0].Name))
+	// s.Fixtures.SQLMock.
+	//	ExpectQuery(regexp.QuoteMeta("SELECT * FROM `enterprises` WHERE name = ? COLLATE NOCASE AND `enterprises`.`deleted_at` IS NULL ORDER BY `enterprises`.`id` LIMIT 1")).
+	//	WithArgs(s.Fixtures.Enterprises[0].Name).
+	//	WillReturnRows(sqlmock.NewRows([]string{"name"}).AddRow(s.Fixtures.Enterprises[0].Name))
 
-	_, err := s.StoreSQLMocked.GetEnterprise(context.Background(), s.Fixtures.Enterprises[0].Name)
+	//_, err := s.StoreSQLMocked.GetEnterprise(context.Background(), s.Fixtures.Enterprises[0].Name)
 
-	s.assertSQLMockExpectations()
-	s.Require().NotNil(err)
-	s.Require().Equal("decrypting secret: failed to decrypt text", err.Error())
+	// s.assertSQLMockExpectations()
+	// s.Require().NotNil(err)
+	// s.Require().Equal("decrypting secret: failed to decrypt text", err.Error())
 }
 
 func (s *EnterpriseTestSuite) TestListEnterprises() {
@@ -266,7 +266,7 @@ func (s *EnterpriseTestSuite) TestListEnterprisesDBFetchErr() {
 
 	s.assertSQLMockExpectations()
 	s.Require().NotNil(err)
-	s.Require().Equal("fetching enterprise from database: fetching user from database mock error", err.Error())
+	s.Require().Equal("fetching enterprises: fetching user from database mock error", err.Error())
 }
 
 func (s *EnterpriseTestSuite) TestDeleteEnterprise() {
@@ -331,7 +331,7 @@ func (s *EnterpriseTestSuite) TestUpdateEnterpriseDBEncryptErr() {
 
 	s.assertSQLMockExpectations()
 	s.Require().NotNil(err)
-	s.Require().Equal("failed to encrypt string", err.Error())
+	s.Require().Equal("encoding secret: invalid passphrase length (expected length 32 characters)", err.Error())
 }
 
 func (s *EnterpriseTestSuite) TestUpdateEnterpriseDBSaveErr() {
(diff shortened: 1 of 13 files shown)


internal/testing/testing.go
```diff
@@ -12,24 +12,20 @@
 //    License for the specific language governing permissions and limitations
 //    under the License.
 
-//go:build testing
-// +build testing
```
matoubidou, line 16: What is the reason for these build constraints?

`_test.go` files are not part of go build - therefore don't go into production code - and dependencies of these files also do not get included? Is there any other reason I miss?

database/sql/enterprise_test.go
```diff
@@ -238,16 +238,16 @@ func (s *EnterpriseTestSuite) TestGetEnterpriseNotFound() {
...
 
-	_, err := s.StoreSQLMocked.GetEnterprise(context.Background(), s.Fixtures.Enterprises[0].Name)
+	//_, err := s.StoreSQLMocked.GetEnterprise(context.Background(), s.Fixtures.Enterprises[0].Name)
 
-	s.assertSQLMockExpectations()
-	s.Require().NotNil(err)
-	s.Require().Equal("decrypting secret: failed to decrypt text", err.Error())
+	// s.assertSQLMockExpectations()
+	// s.Require().NotNil(err)
+	// s.Require().Equal("decrypting secret: failed to decrypt text", err.Error())
```
matoubidou, line 250: The test expects an error for an empty webhook secret in the database. The [validateHookBody](https://github.com/cloudbase/garm/blob/main/runner/runner.go#L544-L547) func seems to imply that empty secrets disables validation. IMHO the implementation is contradictory - this is why I would need your decision on the topic before being able to continue with the implementation

internal/testing/testing.go
gabriel-samfira: This package holds common functions for testing only. These functions can be used in tests in any other package, while any code you define in _test.go are scoped only to the package the tests are in. You cannot import them in tests defined in other packages.

This build constraint makes these functions available only during testing. We don't want to build them otherwise, and suffixing them with _test.go would make them inaccessible to other packages.

You can run:

```bash
make test
```

which sets this tag and runs the tests. If ```vscode``` is complaining, you can set:

```json
"go.buildTags": "testing"
``` 

in your ```settings.json```

database/sql/enterprise_test.go
gabriel-samfira: You are correct. I think we should make signature validation a hard requirement.

gabriel-samfira:
Aligning orgs and enterprises with repos is the desired path in this case. Let's make signature verification a hard requirement. 

Thanks for catching this!

database/sql/enterprise_test.go
matoubidou: Yes, that would have been my decision, too. I think I'll manage to adopt the code beginning of next week. I'll let you know as soon as I am done.

gabriel-samfira: That sounds great! Thanks!

internal/testing/testing.go
matoubidou: Ok... I will drop the commit.

matoubidou:
Hi,

I aligned the code of repositories, organization and enterprises in the `sql` package. I adopted the tests as good as was able to. As I am not familiar with the mocking framework you use, nor understand all the whole testing setup completely it would be great if you could have a close look to the changes I've done.

I am not sure if all these tests are still necessary if we assume that the data already stored in the database is okay, e.g. checking what happens if there is no secret / passphrase stored in the database. Perhaps we just need to make sure that these values go properly to the database. We even might use gorm constraints to guarantee that.

Some thoughts beyond the scope of this PR:

* There is a lot of code duplication due to similar behavior of enterprises, organizations and repositories. Perhaps it would make sense to find a suitable abstraction and type all these things only once.
* In the `SetupTest` funcs of the runner tests I found the fixtures to be initialized with parameters that seem to belong to other entities. E.g. https://github.com/cloudbase/garm/blob/main/runner/enterprise_test.go#L127 refers to `UpdateRepoParams`. I am not sure if this are copy-paste artifacts or if I just do not understand the setup 😕

p1
```diff
@@ -0,0 +1,2 @@
+Identity added: /home/mikuhnt/.ssh/id_rsa (/home/mikuhnt/.ssh/id_rsa)
```
gabriel-samfira, line 1: this is probably not meant to be here :smile: 

There is also a ```garm``` and ```garm-cli``` binary commited.

database/sql/repositories.go
```diff
@@ -27,13 +28,12 @@ import (
...
 
 func (s *sqlDatabase) CreateRepository(ctx context.Context, owner, name, credentialsName, webhookSecret string) (params.Repository, error) {
-	secret := []byte{}
-	var err error
-	if webhookSecret != "" {
-		secret, err = util.Aes256EncodeString(webhookSecret, s.cfg.Passphrase)
-		if err != nil {
-			return params.Repository{}, fmt.Errorf("failed to encrypt string")
-		}
+	if webhookSecret == "" {
```
gabriel-samfira, line 31: This check, and the ```webhookSecret``` checks in the other functions are good, but we should also validate that the create parameters are correct, as early as possible. A good place would be in the ```runner``` package. The ```CreateOrgParams{}```, ```CreateRepoParams{}``` and ```CreateEnterpriseParams{}```, all have a ```Validate()``` function which is called in their respective ```Create()``` functions in the ```runner``` package.

Adding a check there to see if the web hook secret is set, would be nice.

gabriel-samfira (changes requested):
Just a couple of changes are needed, along with adding back the build constraint for the `internal/testing` package, but otherwise looks good.

p1
matoubidou: Sorry.... I rebased mistakenly on another branch. Fixed.

database/sql/repositories.go
matoubidou: I added the checks in the validation funcs. Don't you think that we can remove all other checks later on?

matoubidou:
> Hi @matoubidou !
> 
> I will have a better look at this PR tomorrow. I was considering dropping mocking altogether and just use a DB, but then generating specific error cases would be a bit more difficult. If the only funkiness is the tests, I will gladly merge as is and rectify the tests myself. I am aware they (the tests) need some work.
> 
> I am also aware that there is a lot of code duplication. I would love to allocate some time and attempt to fix that, but that will probably not happen until the end of February.
> 
> To give some context, Garm started out with repositories. Orgs were added after that and in the end we added enterprises. I didn't know how much things would diverge between the different hierarchy levels at the time (especially considering I didn't have access to an enterprise at that time), and wanted to leave some wiggle room in case one of them diverged from the rest in terms of functionality.
> 
> But after adding most of the github interactions, it has become fairly clear where the similarities lie and what is indeed different. I think I would still like to have separate models in the store for repositories, organizations and enterprises, but the code can certainly be de-duped _a lot_.
> 
> Before we cut any kind of release, I would like to have the following as a must:
> 
> * more test coverage
> * de-duplicated code
> 
> A nice to have will be, better scheduling of runners. But that is a big chunk of work which will require a significant refactor of the pool manager. I realize the current state is not ideal in terms of code cleanliness. The important thing is, we are aware of it and will fix it as soon as time permits.

Thanks for the insight. I assume this is how code most of the time evolves. One last nitpick that confused me - the word **runner** for your internal workers could be a little bit misleading for people reading your code (as the purpose of the project is to spin github runners). Perhaps renaming it to **workers** would be great. I will stop with truisms for now 😄.

database/sql/repositories.go
gabriel-samfira: I usually don't do much validation in the store layer on things that should be validated elsewhere. Unless there is an error when trying to encode an empty string, we can remove this check. We need to trust that the data the user wants to save is correct. The store's job is to save and retrieve the data we want. It has no knowledge of the correctness of that data.

So you're right, this can be removed.

gabriel-samfira:
> Thanks for the insight. I assume this is how code most of the time evolves. One last nitpick that confused me - the word **runner** for your internal workers could be a little bit misleading for people reading your code (as the purpose of the project is to spin github runners). Perhaps renaming it to **workers** would be great. I will stop with truisms for now :smile:.

Feel free to nitpick as much as you feel is necessary :smile: . Nothing is set in stone at this point, and a lot of the things were named in haste. Anything that can improve and make this project more useful both for the people using it as well as the people adding code to it is welcome.

It is likely that at some point, garm will be split in multiple components (API, controller/scheduler, workers, etc), allowing us to scale it horizontally when needed. At that point, code will get moved around and decoupled. New names will be given to components. We may rename things and move some code around even before that, but it is a bit low priority for now :smile:.

matoubidou:
Thx for your superfast reaction and merging the issue!

gabriel-samfira:
Thanks for fixing it! :smile:

PR #62, opened 2023-01-19
fix: shortid generates correctly now

maigl:
Sorry...

I committed some un-tested idea where I wanted to remove the ugly '-' and '_' in the shortid.
It doesn't work like that, and currently the runner name generation is broken!

</sub>

### DIFF
--- a/runner/pool/pool.go
+++ b/runner/pool/pool.go
@@ -32,6 +32,7 @@ import (
 	providerCommon "garm/runner/providers/common"
 
 	"github.com/google/go-github/v48/github"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/teris-io/shortid"
 )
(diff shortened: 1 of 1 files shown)


runner/pool/pool.go
```diff
@@ -401,7 +398,7 @@ func (r *basePoolManager) AddRunner(ctx context.Context, poolID string) error {
 	if prefix == "" {
 		prefix = params.DefaultRunnerPrefix
 	}
-	name := fmt.Sprintf("%s-%s", prefix, shortid.MustNew(0, shortidABC, 42).String())
+	name := fmt.Sprintf("%s-%s", prefix, shortid.MustGenerate())
```
gabriel-samfira, line 401: Let's use the

```go
newID, err := shortid.Generate()
if err != nil {
    return errors.Wrap(err, "generating ID")
}
```

We don't want to panic here.

gabriel-samfira:
No worries. Didn't catch the usage of ```Must``` either. I really need to find some time and add more test coverage!

PR #65, opened 2023-01-20
Markdown lint

icadariu:
Hello,
- While reading docs, I've noticed some small mistakes that could be easily fixed using markdown or a spell check.
- Also while trying to check the project, go extension complained about `shortid` which was fixed by running `go mod tidy`.

Thanks,
Ionut


gabriel-samfira:
thanks!

PR #70, opened 2023-01-27
Move some code around

gabriel-samfira:
This change is just a couple of small fixes and moving some code around.

  * metrics are moved to its own package
  * the short ID is sanitized to replace hyphens and dashes with alphanumeric characters
  * the runner prefix is now displayed properly in pool show

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,21 +1,70 @@
-package controllers
+package metrics
 
 import (
 	"log"
+	"sync"
 
 	"garm/auth"
+	"garm/params"
 	"garm/runner"
 
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-type GarmCollector struct {
-	healthMetric   *prometheus.Desc
-	instanceMetric *prometheus.Desc
-	runner         *runner.Runner
+var webhooksReceived *prometheus.CounterVec = nil
+
+// RecordWebhookWithLabels will increment a webhook metric identified by specific
+// values. If metrics are disabled, this function is a noop.
+func RecordWebhookWithLabels(lvs ...string) error {
+	if webhooksReceived == nil {
+		// not registered. Noop
+		return nil
+	}
+
+	counter, err := webhooksReceived.GetMetricWithLabelValues(lvs...)
+	if err != nil {
+		return errors.Wrap(err, "recording metric")
+	}
+	counter.Inc()
+	return nil
 }
 
-func NewGarmCollector(r *runner.Runner) *GarmCollector {
+func RegisterCollectors(runner *runner.Runner) error {
+	if webhooksReceived != nil {
+		// Already registered.
+		return nil
+	}
+
+	garmCollector, err := NewGarmCollector(runner)
+	if err != nil {
+		return errors.Wrap(err, "getting collector")
+	}
+
+	if err := prometheus.Register(garmCollector); err != nil {
+		return errors.Wrap(err, "registering collector")
+	}
+
+	// metric to count total webhooks received
+	// at this point the webhook is not yet authenticated and
+	// we don't know if it's meant for us or not
+	webhooksReceived = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "garm_webhooks_received",
+		Help: "The total number of webhooks received",
+	}, []string{"valid", "reason", "hostname", "controller_id"})
+
+	err = prometheus.Register(webhooksReceived)
+	if err != nil {
+		return errors.Wrap(err, "registering webhooks recv counter")
+	}
+	return nil
+}
+
+func NewGarmCollector(r *runner.Runner) (*GarmCollector, error) {
+	controllerInfo, err := r.GetControllerInfo(auth.GetAdminContext())
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching controller info")
+	}
 	return &GarmCollector{
 		runner: r,
 		instanceMetric: prometheus.NewDesc(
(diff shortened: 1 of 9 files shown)


gabriel-samfira:
@maigl Moved the metrics stuff in a dedicated package. Also made some really small changes here and there. Functionally it should be the same. Let me know if you see any issues with any of this.

gabriel-samfira:
@maigl apropos, the ```webhooksReceived.WithLabelValues()``` function can panic :smile: .Replaced it with ```webhooksReceived.GetMetricWithLabelValues()```.

util/util.go
```diff
@@ -331,3 +333,27 @@ func NewLoggingMiddleware(writer io.Writer) func(http.Handler) http.Handler {
...
+
+func NewID() string {
+	newID, err := shortid.Generate()
+	if err != nil {
+		newID = uuid.New().String()
+	} else {
+		// remove underscores and hyphens from short ID. The hypens will remain
+		// if we are forced to fall back to uuid4.
+		newID = strings.Replace(newID, "_", randomCharacter(), -1)
+		newID = strings.Replace(newID, "-", randomCharacter(), -1)
```
maigl, line 356: I'm no expert here. But I assume that this might compromise the uniqueness of the ids?! You can define the alphabet within the shortid lib, but if you remove '-' and '_' it what other chars instead. 
https://github.com/teris-io/shortid/blob/master/shortid.go#L85 
The algorithm seems to always want 64 distinct chars. But I thought replacing '_' and '-' by '%' and '$' is not much better ..

I really would want to avoid some non-unique issue here.

metrics/metrics.go
```diff
@@ -28,7 +77,28 @@ func NewGarmCollector(r *runner.Runner) *GarmCollector {
...
+
+func (c *GarmCollector) controllerInfo() params.ControllerInfo {
+	controllerInfo, err := c.runner.GetControllerInfo(auth.GetAdminContext())
+	if err != nil {
+		log.Printf("error getting controller info: %s", err)
+		return c.cachedControllerInfo
 	}
+	c.mux.Lock()
+	defer c.mux.Unlock()
+	c.cachedControllerInfo = controllerInfo
```
maigl, line 100: Why is this cached here again? This will only help if runner.GetControllerInfo fails but did work some rounds before, right ?

util/util.go
gabriel-samfira: > I really would want to avoid some non-unique issue here.

indeed. It erred out on OpenStack when the name included an underscore :smile:. The need for 64 unique characters leaves us with few options. We can leave the hyphen and maybe add a ```+``` ? Something that has a lowest chance of failing anywhere garm may be used. Will have to look at the other clouds and what limitations there are in regards to names.

gabriel-samfira: I'll remove this hack as it made my eye twitch as well when I wrote it.

metrics/metrics.go
gabriel-samfira: The ```os.Hostname()``` call can fail, which means we'll end up with no hostname when it does. At least for that call. The cache is to fall back on previous value. What I dislike is that I had to duplicate this in 2 places. I Think I'll add a retry strategy instead and remove the cache.

gabriel-samfira: I'd still like to know when something fails and react accordingly, than to ignore the error.

util/util.go
maigl: Funny.. my openstack is fine with '_'

gabriel-samfira: ohh wait. My bad. The error was on LXD:

```bash
Jan 27 14:18:20 garm garm[11696]: 2023/01/27 14:18:20 failed to add instance to provider: creating instance: creating instance: creating instance: Invalid instance name: Name can only contain alphanumeric and hyphen characters
```

The same thing seems to be true for Azure:

https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules (underscores are not allowed)

From the Azure docs above, in regards to ```VirtualMachines```:

```
Can't use spaces, control characters, or these characters:
~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \ | ; : . ' " , < > / ?

Windows VMs can't include period or end with hyphen.

Linux VMs can't end with period or hyphen.

Which is depressing. Need to find a way to get around this.
```

gabriel-samfira: I think I'll leave the current dictionary alone, and just retry generating a new ID until I get one without those two characters.

gabriel-samfira: actually, just the one. The underscore.

gabriel-samfira: okay. Added a retry (100x). If we can't get a short ID without an underscore in it after 100 tries, we fall back to UUID. Should be fast enough even with this retry.

metrics/metrics.go
gabriel-samfira: Removed the caches and just added a retry in ```GetControllerInfo()```. If it errs there, we just surface that error further along and log it.

util/util.go
maigl: Uh, sorry for being picky now, but trying until no underscore comes is the expensive version of directly replacing it randomly. From an entropy perspective it has the same effect. Should we go back to uuid or maybe a config toggle uuid/shortid?

maigl: Or we use a different library?!

gabriel-samfira: Was looking now for alternatives now. I like the short ID idea.

gabriel-samfira: hmm. I wonder if we could encode the short ID (or even the UUID) into ```base62```. It should solve our problem and shorten even the regular UUID considerably. Uniqueness should be preserved. What do you think?

gabriel-samfira: Looks ok. We get a slightly longer ID as a consequence (between 12 and 14), and the UUID becomes 22 characters long.

maigl: but base 64 contains '/' and '+' and also '=' for padding at the end. 

I also like the shortid though, :smile:

gabriel-samfira: base 62 does not :smile: should be alphanumeric (including upper/lower case)

gabriel-samfira: https://en.wikipedia.org/wiki/Base62

gabriel-samfira: Looks like this:

Screenshot from 2023-01-29 17-51-27

maigl: okay .. never heard of that ;)
sgtm

PR #73, opened 2023-01-31
fix: skip spawn new runners if enough idle runner available

maigl:
Issue:

When we get multiple 'workflow queued' events in a short period of time we seem to have too many instances creating. 
This leads to a situation where we create new instances and scale them down at the same time constantly.

Proposal:
we only scale up if there are not enough free idle_runner

</sub>

### DIFF
--- a/runner/pool/pool.go
+++ b/runner/pool/pool.go
@@ -381,6 +381,24 @@ func (r *basePoolManager) acquireNewInstance(job params.WorkflowJob) error {
 		return nil
 	}
 
+	instances, err := r.store.ListPoolInstances(r.ctx, pool.ID)
+	if err != nil {
+		return errors.Wrap(err, "fetching instances")
+	}
+
+	idleWorkers := 0
+	for _, inst := range instances {
+		if providerCommon.RunnerStatus(inst.RunnerStatus) == providerCommon.RunnerIdle &&
+			providerCommon.InstanceStatus(inst.Status) == providerCommon.InstanceRunning {
+			idleWorkers++
+		}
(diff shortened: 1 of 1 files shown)


gabriel-samfira:
Looks good! Thanks!

gabriel-samfira:
I really need to find some time to rework the way jobs are recorded and runners are spun up. While this does fix situations where pools have a non zero ```MinIdleRunners``` set, it will probably disable pools that don't create idle runners.

After I clear my backlog, I'll look at improving pool management.

maigl:
One solution to this could be to add a 'desired runners' variable which is automatically but slowly adjusted, maybe to the avg parallel active runners of the last half hour or so.. and which can scale between min-idle and max runners.

gabriel-samfira:
As a simple and effective fix for the situation where garm would react to a ```queued``` hook and started to spin up a new runner, clashing with the scale-down feature that was added earlier, we can set a simple grace period on the scale-down function.

Something like:

```go
if time.Since(runner.UpdatedAt).Minutes < 5 {
    // This runner was spun up in the last 5 minutes. Leave it allone for now.
    continue
}
```

And we can remove the check for existing idle runners. The ```queued``` hook fires before an existing idle runner is assigned and starts running the workflow, thus transitioning an idle runner into ```active```, and draining it from the list of available idle runners.

This timeout ensures that a newly spun up idle runner is left alone long enough for it, or one of the older runners to be offered a job by github. So you eliminate that contention between automatically spinning up runners and the scale-down routine killing idle ones.

Issue #76, opened 2023-02-06
Multiple instances with same name on provider


maigl:
As we put more load to garm we ran into a situation where garm created multiple (> 5 time) intances with the very same name on our (openstack) provider.

We think this might be happening here:
https://github.com/cloudbase/garm/blob/main/runner/pool/pool.go#L1045

In case the `setInstanceStatus` fails we simply continue creating the instance and at the same time we don't remember that we already tried to create the same instance. Probably we should use continue to skip creating this instance.
What do you think ?
As the error from `setInstanceStaus` is not logged we currently don't know why it failed.

<sub>Michael Kuhnt <michael.kuhnt@mercedes-benz.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))
</sub>

gabriel-samfira:
Hmm. Garm shouldn't try to create an instance more than once, unless the previous create attempt failed. If it did, it tries to clean the instance here: https://github.com/cloudbase/garm/blob/main/runner/pool/pool.go#L621-L629

We record any error here: https://github.com/cloudbase/garm/blob/main/runner/pool/pool.go#L1053-L1055

We re-queue it here: https://github.com/cloudbase/garm/blob/main/runner/pool/pool.go#L864 up to 5 times.

It may be that the cleanup fails? Do you see any ```failed to cleanup instance``` messages in the log?

gabriel-samfira:
But you are right. We should not re-queue the instance unless cleanup succeeds! Will look into this ASAP and create a PR.

gabriel-samfira:
Could you test out: https://github.com/cloudbase/garm/pull/77. See if this happens again?

maigl:
> Could you test out: #77. See if this happens again?

This runs for a couple of days now .. so far, everything works like a garm .. charm :smile:

gabriel-samfira:
That is awesome!

PR #77, opened 2023-02-06
Bail if we fail to cleanup failed instance

gabriel-samfira:
If we fail to cleanup failed instance, we return before retrying to recreate it.

Fixes: #76 

### DIFF
--- a/runner/pool/pool.go
+++ b/runner/pool/pool.go
@@ -888,6 +888,14 @@ func (r *basePoolManager) retryFailedInstancesForOnePool(pool params.Pool) {
 			// TODO(gabriel-samfira): implement request throttling.
 			if err := r.deleteInstanceFromProvider(inst); err != nil {
 				log.Printf("failed to delete instance %s from provider: %s", inst.Name, err)
+				// Bail here, otherwise we risk creating multiple failing instances, and losing track
+				// of them. If Create instance failed to return a proper provider ID, we rely on the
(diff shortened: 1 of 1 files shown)


maigl:
.. what also could be a related problem is that the whole 'create new runners' and 'delete old runners' process takes longer than the time ticker. This could cause work to pile up.. but I haven't check to code if this is a problem or not.

gabriel-samfira:
> .. what also could be a related problem is that the whole 'create new runners' and 'delete old runners' process takes longer than the time ticker. This could cause work to pile up.. but I haven't check to code if this is a problem or not.

In theory, it should be fine. When we create a new instance we transition from ```pending_create``` to ```creating```. The create loop will ignore anything that is not in ```pending_create```. Same thing for deleting. We transition from ```pending_delete``` to ```deleting```.

runner/pool/pool.go
```diff
@@ -1044,6 +1052,9 @@ func (r *basePoolManager) addPendingInstances() {
 		// won't attempt to create the runner a second time.
 		if err := r.setInstanceStatus(instance.Name, providerCommon.InstanceCreating, nil); err != nil {
 			log.Printf("failed to update runner %s status", instance.Name)
```
maigl, line 1054: let's print the error here:
```suggestion
			log.Printf("failed to update runner %s status: %w", instance.Name, err)
```

Issue #47, opened 2022-10-21
issue with small pool sizes


MoritzKeppler:
A minor issue we found during testing with small pool sizes (in our case: min=0, max=2):
If max pool size is reached garm will ignore additional jobs queued by GH.
As soon as the currently running jobs are completed, the pool is scaled down to min=0 again and the jobs waiting in GH will never be scheduled until we manually resend the webhook event / manually trigger the creation of a new runner.

<sub> Moritz Keppler <moritz.keppler@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

gabriel-samfira:
Indeed. That is a bit of a pickle. Garm creates new runners based on 2 things:

  * an event from github
  * ensuring min_idle_runners

And will obey the max runner limit ignoring "queued" events if that limit is reached. This means that one decision factor is removed in that circumstance. And min_idle_runner=0 will eliminate the other.

An immediate solution would be to have at least 1 set on min_idle_runners. A good solution would be to queue the new runner in the database, and not spin it up until we can accommodate it (number of runners falls under max). Unfortunately, this one will take some time to implement due to backlog. But it will be fixed. 

I for the time being would suggest to define pools at the org/enterprise level (which should reduce the number of pools you will need), and have at least 1 runner per pool.

MoritzKeppler:
thanks! definitely not a critical issue as we have a workaround.

Queued runners in the database could be a solution to at least guarantee that there will be an sufficient amount of runners. But what would happen if the job gets cancelled in the meanwhile or executed on another runner or ...? Maybe easier to have an algorithm that just reacts on the current load, e.g. "always start some extra runners that get automatically shutdown again if not used within a timeframe"... but only a vague idea, not sure if it will work.

gabriel-samfira:
The main issue I see now, is that we ignore events sent via web hooks if ```max_runners``` is reached. That is not a good thing to do, and the issue you're seeing is proof of that.

I think a good approach would be to record web hook events as they come in, and use that as a source of truth to make decisions. It also gives us greater transparency into which event triggered the creation of a runner.

We currently receive web hooks for 3 events:

  * A workflow is ```queued```. This means a workflow has been requested, but no runner picked it up yet.
[...]

PR #86, opened 2023-04-17
Add job tracking

gabriel-samfira:
This branch changes the way garm decides when to spawn a new runner, by implementing some rudimentary job tracking. The bullet points of what is happening are:

  * A new model that tracks jobs has been added
  * When webhooks are sent by github as a result of a job event, we record that job and it's state in the database
  * We no longer create a new runner when a "queued" job shows up, we just record the job
  * When a job transitions to "in_progress", we call ```ensureIdleRunnersForOnePool()``` for the pool to which the runner that has picked up the job, belongs to. Doing so will replace the idle runner if a minimum idle runner value is set to something other than zero, while obeying max runners.
  * A new function called ```consumeQueuedJobs()``` was added. This function will attempt to create a new runner in response to jobs that are in ```queued``` state. It will run before the ```consolidate()``` function. It's purpose is to attempt to consume all ```queued``` jobs by creating any needed runners until the queue is gone. There is a back-off period of 15 seconds before it will attempt to create a new runner. The consolidate loop runs every 5 seconds, so there will be at least ~10 seconds in which an idle runner has time to pick up the job. After that, it will add a new runner.

Some notes:

We cannot control which runners pick up which jobs. We may have pools of runners defined on multiple hierarchy levels that are suitable to sun a particular job. For example, if the job requests a runner with ```self-hosted```, all pools on all levels (repo, org and enterprise) will be able to handle that job.

We may create a runner in pool ```A``` as a response to a ```queued``` job, but a runner from pool ```B``` may pick it up. The runner we spawned may pick up a completely different job than the one that prompted us to create the runner. This does not present an issue as long as all runners were spawned to handle jobs with the same label set. Even if we do end up with some dangling runners, the scale-down logic we have will clean those up after a short period of being idle. While there is no way to guarantee there will be no runner churn, it should be minimal.

For best results, make sure to use unique sets of labels to target runners in a particular pool. Something like this should work:

  * self-hosted
  * x64
  * linux
  * ubuntu-latest
  * gpu

Or if you want to keep it breef, simply use:

  * ubuntu-latest-gpu

If you want to target a particular pool, just add something to narrow down the possibilities just to that pool:

  * ubuntu-latest-gpu
  * ml-team

Fixes: #47
Related: #78 #73


gabriel-samfira:
CC @maigl @MoritzKeppler

maigl:
sounds good  .. we'll check it out, when we find some time .. (maybe not brave enough :)) 
But the new provider currently has higher priority.

gabriel-samfira:
> But the new provider currently has higher priority.

There is no rush at all. Whenever you get a chance. No issues popped up during testing for me, but your use case is more varied than mine. When you do get a chance to test this, ping back.

MoritzKeppler:
first of all: thanks one more time for the PR!

some thoughts:
* wonder if we could run some component test that simulate various scenarios. 
Should be possible to mock GitHub and an external provider; little bit tricky with all the async activities...
* how will the algorithm work if garm is scaled horizontally?
* what will happen if we miss some of the events?

gabriel-samfira:
> first of all: thanks one more time for the PR!
> 
> some thoughts:
> 
>     * wonder if we could run some component test that simulate various scenarios.
>       Should be possible to mock GitHub and an external provider; little bit tricky with all the async activities...

Yup. Should be doable. I need to write proper tests for the whole pool manager.

> 
>     * how will the algorithm work if garm is scaled horizontally?

Sadly, not well. There is currently no global lock. There is a plan to try out etcd as a store, which gives us some nice primitives to work with like watchers and global locks. But I have no ETA on that.

> 
>     * what will happen if we miss some of the events?

They will linker on github until they expire. Sadly, that can't be helped. See this unnecessarily long comment for more details:

https://github.com/cloudbase/garm/pull/86/files#diff-0928304a4f9afb0bf84875a5a19d4d42692d7ed54843569166a233b61cbb7603R1190

Coincidentally, github itself does the same thing. When there is an issue with the gh runners, jobs started while the issue manifested will never be picked up. My guess is that internally they use the same hooks to queue the jobs for their own runners.

maigl:
thx .. will try this out today :)

gabriel-samfira:
> thx .. will try this out today :)

Curious how it works out compared to the old workflow.

PR #87, opened 2023-04-19
Add Userdata options to disable package update during boot / cloud-init

maigl:
Allow to disable an package updates during boot / cloud-init by adding a new field `UserDataOptions` to `BootstrapInstance` parameters.

This does not change the signature of func `GetCloudConfig`. The default behavior does not change.


gabriel-samfira:
Looks good. Let's go with this gor now.

Issue #93, opened 2023-06-04
Set DisableUpdatesOnBoot with LXD?


SystemKeeper:
Maybe I am not seeing it, but is there a way to use `DisableUpdatesOnBoot` (implemented with https://github.com/cloudbase/garm/pull/87) also with LXD? I can change the default in the source, but I am not seeing how I could set the parameter for a specific pool?!

gabriel-samfira:
This is a per-provider setting which is not yet added in LXD. Currently only the external OpenStack provider has it. I can add it to LXD shortly.

SystemKeeper:
>  I can add it to LXD shortly.

That would be great, thank you!

gabriel-samfira:
Grab the latest commit and update your pool as described in the PR. Updates should be skipped for all runners spawned in that pool. Let me know if it works out!

PR #95, opened 2023-06-04
Allow disabling updates

gabriel-samfira:
This change allows disabling updates via extra specs in the LXD provider.

To disable updates, add the following extra specs

```bash
garm-cli pool update \
    --extra-specs='{"disable_updates": true}' \
    <POOL_ID>
```

Fixes: https://github.com/cloudbase/garm/issues/93


SystemKeeper:
Updated and works great, thanks!

PR #97, opened 2023-06-06
Parallelization and LXD timeouts

gabriel-samfira:
* cleanupOrphanedGithubRunners() now uses errgroup to parallelize and report errors when removing runners from the provider.
  * retryFailedInstancesForOnePool() now uses errgroup
  * Removed some setPoolRunningState which should be treated in the loop where those errors eventually bubble up and can be handled.
  * Added a number of timeouts in the LXD provider for delete and list instances. This provider should be converted into an external provider.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/teris-io/shortid v0.0.0-20220617161101-71ec9f2aa569
 	golang.org/x/crypto v0.7.0
 	golang.org/x/oauth2 v0.6.0
+	golang.org/x/sync v0.1.0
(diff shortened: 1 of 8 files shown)


SystemKeeper:
Updated to this PR - let's see 😄

SystemKeeper:
Hey @gabriel-samfira,

looking really good:

```
# lxc list | grep garm- | wc -l
20
``` 

🥳 Not seen any ghost container since this branch!

gabriel-samfira:
awesome! Will merge it. Thanks for testing!

Issue #100, opened 2023-06-08
'garm-cli pool show' hides same values


SystemKeeper:
So we have set `Min Idle runners` and `Max Runners` both to 20. The 20 in `Min Idle Runners` is hidden. When both values are different, both values are shown. 
Before we also had `Runner Bootstrap Timeout` set to 20, in t his case both `Min Idle Runners` and `Runner Bootstrap Timeout` were empty.

Not really a big issue, but it was confusing why the settings were not correctly set at first, but it's just a display thing.

```
# garm-cli pool show f52a63ee-7af9-4a19-b7f0-4f1f67de1579
+--------------------------+----------------------------------------------------------+
| FIELD                    | VALUE                                                    |
+--------------------------+----------------------------------------------------------+
| ID                       | f52a63ee-7af9-4a19-b7f0-4f1f67de1579                     |
| Provider Name            | lxd_local                                                |
| Image                    | gh-ubuntu22-20230507-2.304.0-dockercache                 |
| Flavor                   | runner                                                   |
| OS Type                  | linux                                                    |
| OS Architecture          | amd64                                                    |
| Max Runners              | 20                                                       |
| Min Idle Runners         |                                                          |
| Runner Bootstrap Timeout | 10                                                       |
```

gabriel-samfira:
Ahh yes. This is due to ```AutoMerge``` being set to true in the cli:

https://github.com/cloudbase/garm/blob/main/cmd/garm-cli/cmd/pool.go#L505

I think we can disable ```AutoMerge``` for the second column. The ```AutoMerge``` option prevents printing the same row value on the same column. This is useful for the ```Instances``` row, where we may have multiple instances in a pool, and we don't want to have the first column contain the word ```Instances``` for each printed instance on the second column.

Testing it out.

gabriel-samfira:
This branch should take care of it: https://github.com/cloudbase/garm/pull/101

PR #101, opened 2023-06-08
Don't merge the second column in the CLI

gabriel-samfira:
The second column that is printed when showing a resource, contains values that should not be merged, even if they are similar.

Fixes: https://github.com/cloudbase/garm/issues/100


SystemKeeper:
Thank you 👍

gabriel-samfira:
Thanks for reporting! :smile:

PR #103, opened 2023-06-11
Allow installing additional packages in lxd container

HippocampusGirl:
Dear maintainers,

This proposed change extends the `UserDataOptions` introduced in #87 and #95 so that the user can configure additional packages to be installed into the runner of each pool. 

This can be very useful when GitHub Actions workflows require Docker to be installed in the runner. This can be configured via
```bash
garm-cli pool update \
    --extra-specs='{"extra_packages": ["docker.io"]}' \
        <pool id>
```

### DIFF
--- a/params/params.go
+++ b/params/params.go
@@ -222,7 +222,8 @@ type BootstrapInstance struct {
 }
 
 type UserDataOptions struct {
-	DisableUpdatesOnBoot bool `json:"disable_updates_on_boot"`
+	DisableUpdatesOnBoot bool     `json:"disable_updates_on_boot"`
+	ExtraPackages        []string `json:"extra_packages"`
 }
 
 type Tag struct {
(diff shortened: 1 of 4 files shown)


gabriel-samfira:
This is great! Thanks!

gabriel-samfira:
Apropos, if you need to install packages that are not in the default repositories, you can always have a workflow step that prepares the runner before running anything else:

Here's an example: https://github.com/flatcar/scripts/blob/2f12dfb04bdea32f7627a348c2b24c5f0ec793ed/.github/workflows/ci.yaml#L51-L64

Alternatively, you can [customize an image](https://linuxcontainers.org/lxd/docs/master/howto/images_create/) and use that (saves time installing packages). Or, using the extra specs option you just added. Either way will work :smile:.

Issue #104, opened 2023-06-12
Failed runners are not removed on LXD


SystemKeeper:
Sadly I have to report that it's still not working 100% on LXD.

```
# garm -version
d8ed552

# garm-cli runner list -a
+-------------------+---------+---------------+--------------------------------------+
| NAME              | STATUS  | RUNNER STATUS | POOL ID                              |
+-------------------+---------+---------------+--------------------------------------+
| garm-jWvAZKguDj1R | running | failed        | f52a63ee-7af9-4a19-b7f0-4f1f67de1579 |
+-------------------+---------+---------------+--------------------------------------+
| garm-uzQUmtNIXZyL | running | failed        | f52a63ee-7af9-4a19-b7f0-4f1f67de1579 |
+-------------------+---------+---------------+--------------------------------------+
| garm-H21L1ip0EZvY | running | failed        | f52a63ee-7af9-4a19-b7f0-4f1f67de1579 |
+-------------------+---------+---------------+--------------------------------------+
| garm-GEMOtSLckGH0 | running | failed        | f52a63ee-7af9-4a19-b7f0-4f1f67de1579 |
+-------------------+---------+---------------+--------------------------------------+


# lxc list -c n,s,c,4 | grep garm-jWvAZKguDj1R
| garm-jWvAZKguDj1R | RUNNING | 2023/06/11 15:16 UTC | 172.17.0.1 (docker0) |
# lxc list -c n,s,c,4 | grep garm-uzQUmtNIXZyL
| garm-uzQUmtNIXZyL | RUNNING | 2023/06/11 20:05 UTC | 172.17.0.1 (docker0) |
# lxc list -c n,s,c,4 | grep garm-H21L1ip0EZvY
| garm-H21L1ip0EZvY | RUNNING | 2023/06/12 01:24 UTC | 172.17.0.1 (docker0) |
# lxc list -c n,s,c,4 | grep garm-GEMOtSLckGH0
| garm-GEMOtSLckGH0 | RUNNING | 2023/06/12 02:31 UTC | 172.17.0.1 (docker0) |
```

So running for a while now but never get removed. Garm log has this:

```
# grep GEMOtSLckGH0 *
garm.log:2023/06/12 02:31:31 creating instance garm-GEMOtSLckGH0 in pool f52a63ee-7af9-4a19-b7f0-4f1f67de1579
garm.log:2023/06/12 02:35:02 instance garm-GEMOtSLckGH0 was updated recently, skipping check
[...]

PR #106, opened 2023-06-13
Reap failed agent

gabriel-samfira:
This change reaps runners that appear as `offline` in github and have their `RUNNER STATUS` set to `failed`. Runners that are in this state have registered themselves in GutHub, but failed at a later stage in the setup process. We reap them if no new updates have been made to the instance after the configured pool timeout.

This change also adds a retry when running `config.sh` via userdata. Hopefully in future releases we can switch to using [jit runners](https://github.blog/changelog/2023-06-02-github-actions-just-in-time-self-hosted-runners/) but for the time being, this is not available for GHE.

Fixes: https://github.com/cloudbase/garm/issues/104

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ USER_ID=$(shell ((docker --version | grep -q podman) && echo "0" || id -u))
 USER_GROUP=$(shell ((docker --version | grep -q podman) && echo "0" || id -g))
 ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 GOPATH ?= $(shell go env GOPATH)
+VERSION ?= $(shell git describe --tags --match='v[0-9]*' --dirty --always)
 GO ?= go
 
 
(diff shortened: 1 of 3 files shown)


gabriel-samfira:
I'm going to go ahead and merge this. @SystemKeeper if the problem persists after using this patch, feel free to re-open the issue.

SystemKeeper:
Thanks @gabriel-samfira! I’ll try to install it as soon as possible :)

gabriel-samfira:
No rush! This needed to be done anyway. It should resolve your issue on two fronts:

  * configure is retried 5 times before we give up
  * failed agent is reaped when the pool bootstrap timeout is reached (this is configurable per pool via `garm-cli pool update --runner-bootstrap-timeout`)

So either way, you should see no more runners linger.

gabriel-samfira:
Apropos, this also added a ```make build``` target which just builds the binaries using `go build` and sets the proper version. This is set as a default, so you can just run `make`. The `build-static` target is still there in case you need it.

SystemKeeper:
Quick feedback after this PR: no issues to report 🙂

Issue #113, opened 2023-06-26
Running github runner via sudo - missing environment


SystemKeeper:
As noted in https://github.com/cloudbase/garm/issues/90#issuecomment-1575685747 we try to use an image that is as close as possible to the github action one. We noticed that some workflows fail on the selfhosted runners (but succeed on GitHub), since the executables could not be found. After investigating I noticed that the runner user only has the following in it's PATH environment:
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```

in contrast when manually logging in as `runner` I have this:

```
/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

which is added to `/etc/environment`. 

This happens because the the configuration of the github runner is done with `sudo` and sudo by default replaced the PATH environment as specified in `/etc/sudoers`:

```
# cat /etc/sudoers | grep secure
Defaults	secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```

(the `config.sh` file calls `env.sh` which then saves the PATH inside of the file `.path`)

So question now would be if this is intended or if this behavior should be changed? I see two possibilities here:
1. We change the execution of the `config.sh` file to not use `sudo` and use `su -c '...' runner` instead (would be a change on garm)
2. We keep the current behavior (if it's intended) and change `/etc/sudoers` on our image.

(of course there are more ways as well, lika adjusting the workflows, changing the template, etc., but these were the most obvious ones to me)

Some example outputs:
```
# su --command 'echo $PATH'
/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

sudo -u runner -E cat /proc/self/environ
[...]

PR #115, opened 2023-06-27
Use su to install the runner

gabriel-samfira:
This change executes the runner install script provided via userdata using su. The script itself has also been changed to assume it's running as the runner user.

Fixes: #113

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -277,7 +277,7 @@ func GetCloudConfig(bootstrapParams params.BootstrapInstance, tools github.Runne
 
 		cloudCfg.AddSSHKey(bootstrapParams.SSHKeys...)
 		cloudCfg.AddFile(installScript, "/install_runner.sh", "root:root", "755")
-		cloudCfg.AddRunCmd("/install_runner.sh")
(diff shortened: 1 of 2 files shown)


util/util.go
```diff
@@ -277,7 +277,7 @@ func GetCloudConfig(bootstrapParams params.BootstrapInstance, tools github.Runne
 
 		cloudCfg.AddSSHKey(bootstrapParams.SSHKeys...)
 		cloudCfg.AddFile(installScript, "/install_runner.sh", "root:root", "755")
-		cloudCfg.AddRunCmd("/install_runner.sh")
+		cloudCfg.AddRunCmd(fmt.Sprintf("su -l -c /install_runner.sh %s", appdefaults.DefaultUser))
```
SystemKeeper, line 280: Shouldn't that be the RunnerUsername instead of the default?

SystemKeeper:
Very cool - thanks! Was planning maintenance for tomorrow anyway, perfect timing :-)

util/util.go
gabriel-samfira: Sadly, the `GetCloudConfig()` function just sets the default username at this point. We should make it configurable :smile:.

skjnldsv (approved):

Issue #114, opened 2023-06-27
Support SIGHUP for logrotate


SystemKeeper:
When rotating `garm.log` the file handlers need to be re-opened. Currently we just do a restart of garm when we rotated the log. Would be nice if we could just send a `SIGHUP` (or something else) to gracefully to this without the need to restart.

gabriel-samfira:
Hmm. If you specified the log path specifically and are not logging to syslog, garm should rotate automatically when it reaches ~500 MB or 20 days. See: https://github.com/cloudbase/garm/blob/main/util/util.go#L170-L176

We can also add a `SIGHUP` handler of course.

SystemKeeper:
Oh I see. I wasn't aware of that build-in functionality... Hm, I think SIGHUP wouldn't hurt for people wanting to use logrotate, but I guess the build-in functionality should be be more visible in the future?!

PR #116, opened 2023-06-27
Rotate log file on SIGHUP

ionutbalutoiu:
Fixes https://github.com/cloudbase/garm/issues/114.

Add functionality to rotate the log file when `SIGHUP` signal is received.

Also, a doc is added with few details about logging.

### DIFF
--- a/cmd/garm/main.go
+++ b/cmd/garm/main.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/cloudbase/garm/apiserver/controllers"
@@ -37,6 +38,7 @@ import (
 	"github.com/cloudbase/garm/util"
 	"github.com/cloudbase/garm/util/appdefaults"
 	"github.com/cloudbase/garm/websocket"
+	lumberjack "gopkg.in/natefinch/lumberjack.v2"
 
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
@@ -81,6 +83,26 @@ func main() {
 	if err != nil {
 		log.Fatalf("fetching log writer: %+v", err)
 	}
+
+	// rotate log file on SIGHUP
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, syscall.SIGHUP)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				// Daemon is exiting.
+				return
+			case <-ch:
+				// we got a SIGHUP. Rotate log file.
+				if logger, ok := logWriter.(*lumberjack.Logger); ok {
+					if err := logger.Rotate(); err != nil {
+						log.Printf("failed to rotate log file: %v", err)
+					}
+				}
+			}
+		}
+	}()
 
 	var writers []io.Writer = []io.Writer{
 		logWriter,
--- a/contrib/garm.service
+++ b/contrib/garm.service
@@ -5,6 +5,7 @@ After=multi-user.target
 [Service]
 Type=simple
 ExecStart=/usr/local/bin/garm -config /etc/garm/config.toml
+ExecReload=/bin/kill -HUP $MAINPID
 Restart=always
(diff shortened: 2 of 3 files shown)


util/util.go
```diff
@@ -174,6 +176,16 @@ func GetLoggingWriter(cfg *config.Config) (io.Writer, error) {
 			MaxAge:     28,   // days
 			Compress:   true, // disabled by default
 		}
+		// rotate log file on SIGHUP
```
gabriel-samfira, line 179: This is awesome! But this needs to be in `cmd/garm/main.go`. Any signal handlers need to be handled there.

After we call `GetLoggingWriter()` we can spawn the bellow go routine and wait for a `SIGHUP` signal.  If we have a lumberjack logger, we can call `Rotate()` like you're already doing.

util/util.go
```diff
@@ -174,6 +176,16 @@ func GetLoggingWriter(cfg *config.Config) (io.Writer, error) {
...
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGHUP)
+		go func() {
+			for range ch {
+				if err := writer.(*lumberjack.Logger).Rotate(); err != nil {
+					fmt.Printf("failed to rotate log file: %v\n", err)
```
gabriel-samfira, line 185: You can use `log.Printf()` Even if we're rotating, the log entry will end up somewhere, not on stdout where it may just be lost.

ionutbalutoiu: Yep, after I move the piece of code to `cmd/garm/main.go`, I'll have do that.

ionutbalutoiu: Yep, after I move the piece of code to `cmd/garm/main.go`, I'll do that.

ionutbalutoiu:
@gabriel-samfira review comments addressed.

cmd/garm/main.go
```diff
@@ -82,6 +84,19 @@ func main() {
 		log.Fatalf("fetching log writer: %+v", err)
 	}
 
+	if cfg.Default.LogFile != "" {
+		// rotate log file on SIGHUP
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGHUP)
+		go func() {
+			for range ch {
```
gabriel-samfira, line 92: Try the following:

```go
for {
	select {
		case <-ctx.Done():
			// Daemon is exiting.
			 return
		case <-ch:
			// we got a SIGHUP. Reload logs.
			if err := logWriter.(*lumberjack.Logger).Rotate(); err != nil {
				log.Printf("failed to rotate log file: %v\n", err)
			}
	}
}
```

cmd/garm/main.go
```diff
@@ -82,6 +84,19 @@ func main() {
 		log.Fatalf("fetching log writer: %+v", err)
 	}
 
+	if cfg.Default.LogFile != "" {
```
gabriel-samfira, line 87: You don't need to add this if. We need to install this signal handler regardless, otherwise the daemon will exit on sighup.

ionutbalutoiu: But, if we don't have this `if`, the call:
```go
logWriter.(*lumberjack.Logger).Rotate()
```
from below will fail because `logWriter` will be simply `os.Stdout`, isn't it?

cmd/garm/main.go
```diff
@@ -82,6 +84,19 @@ func main() {
...
+		// rotate log file on SIGHUP
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGHUP)
+		go func() {
+			for range ch {
+				if err := logWriter.(*lumberjack.Logger).Rotate(); err != nil {
```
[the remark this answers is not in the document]
gabriel-samfira: also, you don't need `\n` in the log.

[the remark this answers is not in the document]
ionutbalutoiu: > also, you don't need `\n` in the log.

thanks for mentioning!

gabriel-samfira (approved):

PR #119, opened 2023-06-30
Generate client library via swagger

mihaelabalutoiu:
* Add swagger annotations to `apiserver`
* Generate initial swagger API client code


apiserver/routers/routers.go
```diff
@@ -12,8 +12,24 @@
...
+//	BasePath: /api/v1
+//	Version: 1.0.0
+//	License: Apache 2.0 https://www.apache.org/licenses/LICENSE-2.0
+//
+// swagger:meta
 package routers
 
+//go:generate swagger generate spec --input=../swagger-models.yaml --output=../swagger.yaml --include="routers|controllers"
+//go:generate swagger validate ../swagger.yaml
+//go:generate rm -rf ../client
```
gabriel-samfira, line 30: can we have this live in the top level folder? `client` should be on the same level as `apiserver`

apiserver/routers/routers.go
```diff
@@ -12,8 +12,24 @@
 //    License for the specific language governing permissions and limitations
 //    under the License.
 
+// Package routers Garm API.
+//
+// The Garm API generated using go-swagger.
+//
+//	Schemes: http, https
+//	Host: localhost:9997
```
gabriel-samfira, line 20: Apparently we can [omit the host](https://swagger.io/docs/specification/2-0/api-host-and-base-path/#omit) and have that be filled in dynamically. A potential `codegen` tool or the swagger UI will use whatever host is used to access the definitions.

gabriel-samfira (commented):
This is pretty cool!

If it's still a WiP, feel free to mark it as draft :D

mihaelabalutoiu:
> This is pretty cool!
> 
> If it's still a WiP, feel free to mark it as draft :D

It's no longer WIP. I tested it,  everything is stable.

apiserver/routers/routers.go
mihaelabalutoiu: Ok. Thanks for pointing out.

apiserver/routers/routers.go
mihaelabalutoiu: Sure. Sounds good!

cmd/garm-cli/cmd/repository.go
```diff
@@ -55,17 +56,18 @@ var repoAddCmd = &cobra.Command{
...
-		newRepoReq := params.CreateRepoParams{
+		newRepoReq := apiClientRepos.NewCreateParams()
+		newRepoReq.Body = params.CreateRepoParams{
 			Owner:           repoOwner,
 			Name:            repoName,
 			WebhookSecret:   repoWebhookSecret,
 			CredentialsName: repoCreds,
 		}
-		repo, err := cli.CreateRepository(newRepoReq)
+		response, err := apiCli.Repositories.Create(newRepoReq)
```
gabriel-samfira, line 66: This is great as an example, but we either move everything to the generated client, or we leave everything as it is right now. When we do decide to migrate to the generated client, we need to also change the actual CLI. There is a `--debug` flag that gets passed to `resty` that prints the raw response. This is just an example, but a change to the actual CLI needs to be cohesive.

gabriel-samfira: To clarify, we can still have the generated client library live in this repo. But the actual CLI needs to be migrated later to use it. Until then, we can skip the change in this file.

ionutbalutoiu: > This is great as an example, but we either move everything to the generated client, or we leave everything as it is right now. When we do decide to migrate to the generated client, we need to also change the actual CLI. There is a `--debug` flag that gets passed to `resty` that prints the raw response. This is just an example, but a change to the actual CLI needs to be cohesive.

https://github.com/cloudbase/garm/pull/119/files#diff-5014f3eeef4459bc7cc5b8f7cc8b57ac102b493d2aa20696e82feb1376ee1310R92

^ this does the trick to have the same output of the commands with `--debug` as we have with `resty`:
```go
if debug {
	os.Setenv("SWAGGER_DEBUG", "true")
}
```

this will make the swagger CLI autogenerated library to use debug mode on:
https://github.com/go-openapi/runtime/blob/1ade6d4fa0ddfcfdcad57761eed3588b40cf41ae/logger/logger.go#L11-L19

and the output of the `gram-cli --debug ...` is similar to what we have with `resty` debug enabled.

ionutbalutoiu: > To clarify, we can still have the generated client library live in this repo. But the actual CLI needs to be migrated later to use it. Until then, we can skip the change in this file.

but this makes sense to me! I agree with it.

mihaelabalutoiu:
@gabriel-samfira  I just addressed the review comments. Please take a look again. Thank you!

gabriel-samfira:
Awesome! Thanks!

PR #145, opened 2023-07-27
Add GitHub workflow for integration tests

mihaelabalutoiu:
* This PR adds `GitHub workflow` that runs `integration tests`.

* The steps in the integration tests are:
  *  `Setup Golang`
  *  `Setup LXD`
  *  `Build GARM`
  *  `Setup GARM `
  *  `Generate secrets`
  *  `Setup ngrok`
  *  `Run integration tests`
  *  `Show GARM logs`

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,51 @@
+name: Integration Tests
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Golang
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+
+      - name: Setup LXD
+        uses: canonical/setup-lxd@v0.1.1
+
+      - name: Build GARM
+        run: make build
+
+      - name: Setup GARM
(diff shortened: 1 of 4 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,174 @@
...
+        run: |
+          set -e
+          set -x
+          set -o pipefail
+
+          GARM_URL="http://127.0.0.1:9997"
+          ADMIN_USERNAME='admin'
+          ADMIN_PASWORD='Ma*jirksiCr9esrT8DGQPrfx*XKnBVZo'
+
+          garm-cli init --debug \
```
gabriel-samfira, line 134: While it's great that the CLI gets some tests as well, I think that it would be better if you used the code you already have here:

https://github.com/mihaelabalutoiu/garm-test-client/blob/main/main.go

We can set up a nice integration test suite, that spins up the GARM API and daemon loops, then run the code at the above link to test everything.

This way we can easily see if the output from the API matches expectations. We also end up testing both GARM and the client library.

The CLI is just a wrapper on top of the client library anyway (with some pretty tables).

gabriel-samfira: we can also just set up garm as a service like you did above and run the code at the link against it. Both options are fine.

mihaelabalutoiu: Sounds good! I will setup `garm` as a service and run that code.  Thank you :)

mihaelabalutoiu:
Heya @gabriel-samfira I updated the PR, please take a look now. Thank you :)

PR #149, opened 2023-08-07
Cleaning up leftover runners for `org/repo`

mihaelabalutoiu:
* This PR implements cleanup of the `orphan runners` for `org/repo`.
* I updated the PR with the following changes:
  * I replaced the `log.Fatalf` with `panic` to prevent the `exit 1` from interrupting the defer statement.
  * I put `graceful cleanup functions` in a function called `GracefulCleanup` and deferred it first.
  * Make sure that the `graceful cleanup functions` are idempotent.
  * I also added two functions: (`GhOrgRunnersCleanup` and `GhRepoRunnersCleanup`) to forcefully delete the runners via  `GitHub API` and deferred them after the `GracefulCleanup` function.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -72,6 +72,7 @@ jobs:
           ORG_NAME: gsamfira
           REPO_NAME: garm-testing
           CREDENTIALS_NAME: test-garm-creds
+          GH_TOKEN: ${{ secrets.GH_OAUTH_TOKEN }}
 
       - name: Show GARM logs
         if: always()
--- a/test/integration/e2e.go
+++ b/test/integration/e2e.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -24,6 +25,8 @@ import (
 	"github.com/cloudbase/garm/params"
 	"github.com/go-openapi/runtime"
 	openapiRuntimeClient "github.com/go-openapi/runtime/client"
+	"github.com/google/go-github/v53/github"
+	"golang.org/x/oauth2"
 )
 
 var (
@@ -51,6 +54,7 @@ var (
 	email    = os.Getenv("GARM_EMAIL")
 	name     = os.Getenv("GARM_NAME")
 	baseURL  = os.Getenv("GARM_BASE_URL")
+	ghtoken  = os.Getenv("GH_TOKEN")
 
 	poolID string
 )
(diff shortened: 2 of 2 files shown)


test/integration/scripts/cleanup.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
gabriel-samfira, line 1: If garm is still running when you clean up runners and has a min_idle_runners set, it may clean up the orphaned runners then spin up a new one. 

The proper way to clean would be:

* garm-cli pool update --enabled=false <pool ID>
* garm-cli runner rm -f <runner name>

You can loop through all pools and disable them, then loop through any existing runners and remove them. You don't need to wait for the runner to dissapear from GARM, as the VM with the runners will be torn down, and because once `garm-cli runner rm` returns, the runner will be removed from github.

ionutbalutoiu: I believe we can simply stop GARM systemd service, and have this code executed.

What do you think?

gabriel-samfira: Yes and no. This clobbers ALL runners, even those that belong to other garm controllers. If multiple instances of this test runs, or if we attach another garm controller to the same repo/org, this test will remove all runners, including those that don't belong to the controller that ran in this test.

The `garm-cli runner rm -f` command will only remove runners spun up by this test.

ionutbalutoiu: Fair point. This cleanup assumes that we don't have concurrent jobs.

I'm hesitant on relying on GARM being responsive, since we might have potential regressions during the dev cycle, and cleanup would fail. In that scenario, we would have to manually cleanup runners from jobs that failed the cleanup.

However, we can rely on `gh` to be stable enough to cleanup the runners.

I believe it would be perfect to save the controller ID after GARM init, and we can use for filtering here.

ionutbalutoiu: As a side note, the integration tests contain a part of "graceful cleanup" at the end of the job:
https://github.com/cloudbase/garm/blob/1a07073820323302295a2c7cda95a87ffd5cc1b3/test/integration/e2e.go#L1056-L1060

So, we don't avoid testing those API's from GARM.

gabriel-samfira: These should be deferred to always run.

We can use the `gh` cli, but lets filter runners by controller tag.

ionutbalutoiu: > These should be deferred to always run.
> 
> We can use the `gh` cli, but lets filter runners by controller tag.

That's a good point. Let's have these deferred, and use `gh` to cleanup filtered runners by controller tag.

ionutbalutoiu: I just realized that the `FirstRunHandler` from the `/api/v1/first-run` returns a `params.User` struct on successful response:
https://github.com/cloudbase/garm/blob/1a07073820323302295a2c7cda95a87ffd5cc1b3/apiserver/controllers/controllers.go#L278-L316

I looked in the GARM API, and there seems to be no way to get the GARM controller ID from the client perspective. It seems to be used only internally in the GARM server.

But, we can have the integration tests save all the pool IDs that were created, and we filter by pool ID here.

gabriel-samfira: pool IDs also work. Worst case, the cleanup step can always interogate the DB. But the pool IDs are enough.

mihaelabalutoiu: Hi @gabriel-samfira , I have made the changes. Please take a look now and revisit the description as well. Thank you! : )

PR #154, opened 2023-08-15
Add webhook management for repositories and organizations

gabriel-samfira:
This change adds the ability to install or uninstall the GitHub web hook in repositories and organizations (no enterprise support yet) that is needed by GARM to operate normally.

Currently, users have to manually configure a web hook in GitHub that points back to GARM. This is needed in order for garm to receive notifications about jobs, and react by creating or removing runners. In the current implementation, the API endpoint for webhooks is `/webhooks`.

This change makes it possible for GARM to also receive webhooks on `/webhooks/{controllerID}`, where `{controllerID}` is the `UUID4` of the current controller.

The old `/webhook` endpoint still works and can be used for manual configuration. When installing or removing webhooks, GARM will only use the API endpoint that includes the controller ID. This way, we don't accidentally remove a user-defined web-hook.

Or course, users can also manually configure web-hooks using the controllerID suffixed web-hook API endpoint if they wish, but in that case, GARM may remove the webhook if requested via CLI.

To use this feature, the PAT associated to a repo or org needs to have access to add/remove webhooks. A new setting will also be added that enables the API endpoints needed to manage web hooks. 

Example usage:

You will have to add the webhook base URL in your `config.toml` file under the `[DEFAULT]` section and enable the webhook management feature:

```toml
[DEFAULT]
webhook_url = "https://garm.example.com/webhooks"
enable_webhook_management = true
```
Then you can install/uninstall webhooks.

Installing the webhook:

```bash
garm-cli repo webhook install <REPO_ID>
```

Installing the webhook and disabling TLS certificate verification (not recommended):

```bash
garm-cli repo webhook install --insecure <REPO_ID>
```

Uninstalling:

```bash
garm-cli repo webhook uninstall <REPO_ID>
```

Show current status:

```bash
ubuntu@garm:~/garm$ garm-cli org webhook show 23ff4c81-b998-42cf-a4d5-f0c095f3dc4f
+--------------+----------------------------------------------------------------------------+
| FIELD        | VALUE                                                                      |
+--------------+----------------------------------------------------------------------------+
| ID           | 429050643                                                                  |
| URL          | https://garm.example.com/webhooks/543d1288-29de-41b5-8200-f9d3f20502d5     |
| Events       | [workflow_job]                                                             |
| Active       | true                                                                       |
| Insecure SSL | false                                                                      |
+--------------+----------------------------------------------------------------------------+

```

To view your current controller info and configured URLs, you can run:

```bash
ubuntu@garm:~/garm$ garm-cli controller-info show
+------------------------+----------------------------------------------------------------------------+
| FIELD                  | VALUE                                                                      |
+------------------------+----------------------------------------------------------------------------+
| Controller ID          | 543d1288-29de-41b5-8200-f9d3f20502d5                                       |
| Hostname               | garm                                                                       |
| Metadata URL           | https://garm.example.com/api/v1/metadata                                   |
| Callback URL           | https://garm.example.com/api/v1/callbacks                                  |
| Webhook URL Base       | https://garm.example.com/webhooks                                          |
| Controller Webhook URL | https://garm.example.com/webhooks/543d1288-29de-41b5-8200-f9d3f20502d5     |
+------------------------+----------------------------------------------------------------------------+
```

## Important notes

When removing a repo or an organization, by default garm will also cleanup any webhooks that may have been created. GARM will only ever touch webhooks that exactly match both the base URL configured in `webhook_url` **and** the `controllerID` suffix.

IF we assume that the controller ID that was automatically generated when you initialized GARM is `5eb2fc8d-b69b-4690-8e09-1c813b99bbf5` and your webhook URL is `https://garm.example.com/webhooks`, then GARM will only cleanup webhooks that exactly match `https://garm.example.com/webhooks/5eb2fc8d-b69b-4690-8e09-1c813b99bbf5`.

This behavior can be disabled by adding `?keepWebhook=1` to the `DELETE` request or by using the `--keep-webhook=true` flag for the `org delete` and `repo delete` CLI commands.


gabriel-samfira:
Merging as is for now. Integration tests are forthcoming.

PR #180, opened 2023-10-06
feat: add new metrics

bavarianbidi:
add info metrics about providers, enterprises, organizations, repositories and pools.

Also expose most of the configurable pool information as metric like e.g. max Runners as garm_pool_max_runners

### DIFF
--- a/metrics/enterprise.go
+++ b/metrics/enterprise.go
@@ -1,0 +1,50 @@
+package metrics
+
+import (
+	"log"
+	"strconv"
+
+	"github.com/cloudbase/garm/auth"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// CollectOrganizationMetric collects the metrics for the enterprise objects
+func (c *GarmCollector) CollectEnterpriseMetric(ch chan<- prometheus.Metric, hostname string, controllerID string) {
+	ctx := auth.GetAdminContext()
+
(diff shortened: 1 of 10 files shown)


metrics/enterprise.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// CollectOrganizationMetric collects the metrics for the enterprise objects
+func (c *GarmCollector) CollectEnterpriseMetric(ch chan<- prometheus.Metric, hostname string, controllerID string) {
+	ctx := auth.GetAdminContext()
+
+	enterprises, err := c.runner.ListEnterprises(ctx)
+	if err != nil {
+		log.Printf("listing providers: %s", err)
+		// continue anyway
```
gabriel-samfira, line 18: in all likelihood the `enterprises` variable is an empty list, so you can simply `return` here and skip the rest of the code.

gabriel-samfira (changes requested):
This looks great! There is one small nit that happens in most collectors.  You can replace `// continue anyway` with a simple return and get the same effect.

metrics/enterprise.go
bavarianbidi: found this pattern in the existing code and I just kept using it without thinking :see_no_evil:

gabriel-samfira: No worries! There are plenty of places we can improve in time.

PR #188, opened 2023-11-30
feat: passthrough additional env vars to provider bin

### LINKED ISSUE #187: external Provider: allow external provider configuration via environment variables
As an external provider developer, i want to make use of environment variables to configure the provider binary itself.

**current state:**

When `garm` calls the external provider binary, only certain environment variables are passed into the provider execution

https://github.com/cloudbase/garm/blob/05e179604dedeada5354495ab319633fb444a0ab/runner/providers/external/external.go#L69-L86

**problem statement:**

We are implementing a kubernetes based provider where it should be possible to have the kubernetes based pod runners within the same cluster where `garm` itself is running (mainly for development purpose).
For that we have to create a `kubernetes` go-client. By using the official `client-go` package it's possible to "discover" if the client must be generated via a given `kubeconfig` or if the client could get generated from the information it can "auto-discover" via environment variables (if `KUBERNETES_SERVICE_PORT` and `KUBERNETES_SERVICE_HOST` are set, the assumption is, that we are running inside kubernetes - https://github.com/kubernetes/client-go/blob/v0.28.4/rest/config.go#L507-L541)

**feature request:**

The provider-configuration should have a list of environment variables, which are passed through `garm` into the provider-binary execution.

Note: would be nice to have both, a list of environment variables and some kind of prefix based envs (e.g. `KUBERNETES_*`)

bavarianbidi:
as some provider binaries probably need additional environment variables set (e.g kubernetes as client-go depends on KUBERNETES_SERVICE_ vars) it should be possible to define a list of environment variables which should get bypassed into the provider binary execution

tested with our kubernetes-provider binary where i want to bypass the `KUBERNETES_` variables into the binary execution call

```toml
[[provider]]
name = "kubernetes_external"
description = "kubernetes provider"
provider_type = "external"
[provider.external]
config_file = "/home/mb/bin/garm/garm-provider-k8s-config.yaml"
provider_executable = "/home/mb/bin/garm/garm-provider-k8s"
environment_variables = [
  "KUBERNETES_SERVICE_PORT",
  "KUBERNETES_SERVICE_HOST",
  "KUBERNETES_",
]
```

Fixes #187


config/external.go
```diff
@@ -42,6 +43,25 @@ type External struct {
 	// the provider. If specified, it will take precedence over the "garm-external-provider"
 	// executable in the ProviderDir.
 	ProviderExecutable string `toml:"provider_executable" json:"provider-executable"`
+	// EnvironmentVariables is a map of environment variable names that will be
```
gabriel-samfira, line 46: nit: this seems to be an array, not a map :smile:

runner/providers/external/external.go
```diff
@@ -74,6 +79,7 @@ func (e *external) CreateInstance(ctx context.Context, bootstrapParams commonPar
 		fmt.Sprintf("GARM_POOL_ID=%s", bootstrapParams.PoolID),
 		fmt.Sprintf("GARM_PROVIDER_CONFIG_FILE=%s", e.cfg.External.ConfigFile),
 	}
+	asEnv = append(asEnv, e.environmentVariables...)
```
gabriel-samfira, line 82: We should probably dedup the variables. We also need to make sure that the values in `asEnv` take precedence to those in the env. We don't want to accidentally override `GARM_COMMAND` for example.

I know we have an include list of allowed environment variables, so unless we explicitly set `GARM_` in the config to allow those variables to be passed through, there is no chance of collision, but it may not be a bad idea to account for mistakes.

Or we can skip any pattern set in the config that starts with `GARM`, ensuring that we can't accidentally override the variables set by GARM.

config/external.go
```diff
@@ -42,6 +43,25 @@ type External struct {
...
+}
+
+func (e *External) GetEnvironmentVariables() []string {
+	envVars := []string{}
+
+	for _, configuredEnvVars := range e.EnvironmentVariables {
+		// discover environment variables
+		for _, k := range os.Environ() {
+			variable := strings.SplitN(k, "=", 2)
+
```
gabriel-samfira, line 58: We should probably skip any pattern that starts with `GARM`. We don't want to override the ones that are set by garm. See bellow comment.

gabriel-samfira (changes requested):
This looks great! Just one meaningful comment in regards to skipping any pattern that starts with `GARM`, to protect the variables that should be set by GARM itself when calling the external provider binary.

runner/providers/external/external.go
bavarianbidi: sounds totally valid - thanks for the finding. Will skip `GARM_` prefixed variables

config/external_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
```
gabriel-samfira, line 21: This package was [introduced in go version `1.21.4`](https://pkg.go.dev/slices). We need to change `go.mod`:

https://github.com/cloudbase/garm/blob/main/go.mod#L3

gabriel-samfira (approved):

PR #189, opened 2023-12-11
Add option to disable JIT config

gabriel-samfira:
This change adds a flag on providers that allows users to disable JIT configuration even when it's available. For context, JIT is available on github.com and any GHES instance >=3.10.

This option is a stopgap measure for providers that have not yet been updated to use JIT configs instead of runner registration tokens.


gabriel-samfira:
CC @bavarianbidi

bavarianbidi:
thank your for the fast fix!

PR #198, opened 2023-12-18
Create needed folders before use

gabriel-samfira:


gabriel-samfira:
In hindsight, I should have probably tested before creating 4 PRs that fix a bunch of "oops" errors.

PR #215, opened 2024-02-16
Dockerfile: Build actually static binaries

jepio:
Without '-linkmode external' some of the produced binaries were not actually static. /bin/garm was fine but providers would fail with "exit code 127" when executed. Unpacking the container image and running the binaries in strace showed that they were accessing the musl dynamic linker.


gabriel-samfira (approved):

gabriel-samfira:
Thanks!

gabriel-samfira:
Running the image build action now. It should finish in about 30 minutes.

Issue #181, opened 2023-10-06
extend existing metrics endpoint


bavarianbidi:
I want to get know how often the runner creation failed as the provider-plugin reported an error.

With the existing `/metrics` endpoint we should expose the following information:

```
garm_runner_created_count{provider="provider_name"}
garm_runner_failed_count{provider="provider_name"}
```

With that it should be easy to visualize (and possible alert) on a certain rate.

Additional information:
To differentiate the failures, we could/should check if `ProviderFault` could be also used as information which we reflect in the metrics via a `provider_fault` label (https://github.com/cloudbase/garm/blob/b1dd54c07e989d5335cd864cedb0efcbfee74dc0/vendor/github.com/cloudbase/garm-provider-common/params/params.go#L152-L154)

bavarianbidi:
As i'm not able to assign the issue to myself: i currently work on this.

bavarianbidi:
Following the pattern from [here](https://promlabs.com/blog/2023/09/19/errors-successes-totals-which-metrics-should-i-expose-to-prometheus/#recommended-for-binary-outcomes-exposing-errors-and-totals) i currently think about the following additional counters - just a rough idea:

* `instance_created_total`
* `instance_created_errors_total`
* `github_api_runner_delete_total`
* `github_api_runner_delete_errors_total`
* `pool_update_total`
* `pool_update_errors_total`

bavarianbidi:
as all the needed/wanted metrics got added with https://github.com/cloudbase/garm/pull/217 i will close this issue

PR #216, opened 2024-02-19
chore: refactor metrics endpoint

bavarianbidi:
refactoring is needed to make the metrics package usable from within the runner package for further metrics.

This change also makes the metric-collector independent from requests to the /metrics endpoint

This refactoring is needed to continue on #181

### DIFF
--- a/cmd/garm/main.go
+++ b/cmd/garm/main.go
@@ -35,8 +35,8 @@ import (
 	"github.com/cloudbase/garm/config"
 	"github.com/cloudbase/garm/database"
 	"github.com/cloudbase/garm/database/common"
-	"github.com/cloudbase/garm/metrics"
 	"github.com/cloudbase/garm/runner"
+	runnerMetrics "github.com/cloudbase/garm/runner/metrics"
 	garmUtil "github.com/cloudbase/garm/util"
 	"github.com/cloudbase/garm/util/appdefaults"
 	"github.com/cloudbase/garm/websocket"
@@ -214,13 +214,13 @@ func main() {
 
 	router := routers.NewAPIRouter(controller, jwtMiddleware, initMiddleware, instanceMiddleware, cfg.Default.EnableWebhookManagement)
 
+	// start the metrics collector
 	if cfg.Metrics.Enable {
-		slog.InfoContext(ctx, "registering prometheus metrics collectors")
-		if err := metrics.RegisterCollectors(runner); err != nil {
-			log.Fatal(err)
-		}
 		slog.InfoContext(ctx, "setting up metric routes")
 		router = routers.WithMetricsRouter(router, cfg.Metrics.DisableAuth, metricsMiddleware)
+
+		slog.InfoContext(ctx, "start metrics collection")
+		runnerMetrics.CollectObjectMetric(runner, time.NewTicker(cfg.Metrics.Period))
 	}
 
 	if cfg.Default.DebugServer {
--- a/config/config.go
+++ b/config/config.go
@@ -456,8 +456,9 @@ func (t *TLSConfig) Validate() error {
 }
 
 type Metrics struct {
-	DisableAuth bool `toml:"disable_auth" json:"disable-auth"`
-	Enable      bool `toml:"enable" json:"enable"`
+	DisableAuth bool          `toml:"disable_auth" json:"disable-auth"`
+	Enable      bool          `toml:"enable" json:"enable"`
+	Period      time.Duration `toml:"period" json:"period"`
 }
 
 // APIServer holds configuration for the API server
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,206 +1,41 @@
 package metrics
 
 import (
-	"log/slog"
-
-	"github.com/cloudbase/garm/auth"
-	"github.com/cloudbase/garm/params"
-	"github.com/cloudbase/garm/runner"
-
-	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const metricsNamespace = "garm_"
-const metricsRunnerSubsystem = "runner_"
-const metricsPoolSubsystem = "pool_"
-const metricsProviderSubsystem = "provider_"
-const metricsOrganizationSubsystem = "organization_"
-const metricsRepositorySubsystem = "repository_"
-const metricsEnterpriseSubsystem = "enterprise_"
-const metricsWebhookSubsystem = "webhook_"
+const metricsNamespace = "garm"
+const metricsRunnerSubsystem = "runner"
+const metricsPoolSubsystem = "pool"
+const metricsProviderSubsystem = "provider"
+const metricsOrganizationSubsystem = "organization"
+const metricsRepositorySubsystem = "repository"
+const metricsEnterpriseSubsystem = "enterprise"
+const metricsWebhookSubsystem = "webhook"
 
-var webhooksReceived *prometheus.CounterVec = nil
-
-// RecordWebhookWithLabels will increment a webhook metric identified by specific
-// values. If metrics are disabled, this function is a noop.
-func RecordWebhookWithLabels(lvs ...string) error {
-	if webhooksReceived == nil {
-		// not registered. Noop
-		return nil
-	}
-
-	counter, err := webhooksReceived.GetMetricWithLabelValues(lvs...)
-	if err != nil {
-		return errors.Wrap(err, "recording metric")
-	}
-	counter.Inc()
-	return nil
-}
-
-func RegisterCollectors(runner *runner.Runner) error {
-	if webhooksReceived != nil {
-		// Already registered.
-		return nil
-	}
-
-	garmCollector, err := NewGarmCollector(runner)
-	if err != nil {
-		return errors.Wrap(err, "getting collector")
-	}
-
-	if err := prometheus.Register(garmCollector); err != nil {
-		return errors.Wrap(err, "registering collector")
-	}
-
-	// metric to count total webhooks received
-	// at this point the webhook is not yet authenticated and
-	// we don't know if it's meant for us or not
-	webhooksReceived = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Name: metricsNamespace + metricsWebhookSubsystem + "received",
-		Help: "The total number of webhooks received",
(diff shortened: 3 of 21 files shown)


bavarianbidi:
@gabriel-samfira by refactoring all the metrics, i've seen that we sometimes add the labels `hostname` and `controller_id` (by using `r.GetControllerInfo(ctx)`) to the defined metrics. But sometimes we don't add this.

Will do another round in this PR and align all the labels to have at least the `controller_id` as label. We do not need the `hostname` label, but if you prefer to have it, please let me know.

gabriel-samfira:
This is great! As part of the next release, I was thinking of extending the external provider interface to allow an RPC API of some sort that external providers can call into for logging, telemetry, etc. We can also do metrics there once that happens, like how long executing an operation against a provider took and things of that sort.

Will do an in-depth review soon.

runner/metrics/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"log/slog"
+	"time"
+
+	"github.com/cloudbase/garm/auth"
+	"github.com/cloudbase/garm/runner"
+)
+
+func CollectObjectMetric(r *runner.Runner, timer *time.Ticker) {
```
gabriel-samfira, line 11: This function should accept a context. Then we can pass in the global context that gets generated in `main()`. The reason is that we need to end the goroutine bellow when `ctx.Done()` returns.

runner/metrics/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	ctx := auth.GetAdminContext()
+
+	controllerInfo, err := r.GetControllerInfo(ctx)
+	if err != nil {
+		slog.With(slog.Any("error", err)).ErrorContext(ctx, "cannot get controller info")
+	}
+
+	go func() {
+		// we wan't to initiate the collection immediately
+		for ; true; <-timer.C {
```
gabriel-samfira, line 22: This should probably have a `select{}` block and return if `ctx.Done()` returns.

runner/metrics/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	"log/slog"
+	"time"
+
+	"github.com/cloudbase/garm/auth"
+	"github.com/cloudbase/garm/runner"
+)
+
+func CollectObjectMetric(r *runner.Runner, timer *time.Ticker) {
+
+	ctx := auth.GetAdminContext()
```
gabriel-samfira, line 13: This could accept a new parameter of type `context.Context` and return a new context with the appropriate admin fields set. If `nil` is passed, it can generate a `context.Background()` as it does now.

gabriel-samfira (changes requested):
I love the separation of concerns and how much cleaner the code is. I have just a couple of comments, but overall this looks much nicer!

runner/metrics/metrics.go
bavarianbidi: done with https://github.com/cloudbase/garm/pull/216/commits/97f172eb513bda47d6658b81fd9bf262d687aa98

runner/metrics/metrics.go
bavarianbidi: done with https://github.com/cloudbase/garm/pull/216/commits/97f172eb513bda47d6658b81fd9bf262d687aa98

runner/metrics/metrics.go
bavarianbidi: done with https://github.com/cloudbase/garm/pull/216/commits/97f172eb513bda47d6658b81fd9bf262d687aa98

config/config.go
```diff
@@ -456,8 +456,39 @@ func (t *TLSConfig) Validate() error {
...
+	duration, err := time.ParseDuration(fmt.Sprint(m.Period))
+	if err != nil {
+		return 0, err
+	}
+	return duration, nil
+}
+
+// Duration returns the configured duration or the default duration if no value
+// is configured or the configured value is invalid.
+func (m *Metrics) Duration() time.Duration {
```
gabriel-samfira, line 482: This is great!

gabriel-samfira (changes requested):
Just a couple of really tiny suggestions to align with the rest of the code base. Overall this looks great! Thanks for all the work!

runner/metrics/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+	ctx = auth.GetAdminContext(ctx)
+
+	controllerInfo, err := r.GetControllerInfo(ctx)
+	if err != nil {
+		slog.With(slog.Any("error", err)).ErrorContext(ctx, "cannot get controller info")
+	}
+
+	go func() {
+		// we wan't to initiate the collection immediately
+		for ; true; <-ticker.C {
```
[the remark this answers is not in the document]
bavarianbidi: i wanted to have the metrics data collection started imidiatly and not wait for the first `ticker.C`
That's why `for ; true; <-ticker.C {` and the `default` section in the `select` statement.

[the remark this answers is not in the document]
bavarianbidi: @gabriel-samfira 
i wanted to have the metrics data collection started imidiatly and not wait for the first `ticker.C`
That's why `for ; true; <-ticker.C {` and the `default` section in the `select` statement.

metrics/metrics.go
```diff
@@ -1,206 +1,41 @@
...
+const metricsPoolSubsystem = "pool"
+const metricsProviderSubsystem = "provider"
+const metricsOrganizationSubsystem = "organization"
+const metricsRepositorySubsystem = "repository"
+const metricsEnterpriseSubsystem = "enterprise"
+const metricsWebhookSubsystem = "webhook"
+
+func init() {
+	// runner metrics
+	prometheus.MustRegister(InstanceStatus)
```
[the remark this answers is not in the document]
bavarianbidi: will take a look on that - maybe there is something from the prometheus package we can re-use :eyes:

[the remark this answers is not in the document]
gabriel-samfira: Usually, where there's a `MustSomething()` there is also a `Something() error` function. See:

https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Register

runner/metrics/metrics.go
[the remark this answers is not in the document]
gabriel-samfira: Ahh. Another solution would be to split out all those metrics calls into a function, call it once before the for loop and call it again in the ticker case.

Basically everything in the `default` branch can be put into a private function:

```go
func collectAllMetrics(ctx context.Context) error {
  return nil
}
```

Then right before the for loop you can call it once:

```go
if err := collectAllMetrics(ctx); err != nil {
  // log the error
}
```

Then again in the ticker case.

metrics/metrics.go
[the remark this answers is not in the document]
bavarianbidi: done in https://github.com/cloudbase/garm/pull/216/commits/17d74dfbf00a561cee17272905a0abe8e35c5a7b

runner/metrics/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"log/slog"
+	"time"
+
+	"github.com/cloudbase/garm/auth"
+	"github.com/cloudbase/garm/runner"
+)
+
+func CollectObjectMetric(ctx context.Context, r *runner.Runner, ticker *time.Ticker) {
```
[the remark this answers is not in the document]
bavarianbidi: https://github.com/cloudbase/garm/pull/216/commits/0a53b8f6d8635ae1c03d871349566e9df7cb5b5d

runner/metrics/metrics.go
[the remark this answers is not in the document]
bavarianbidi: done in https://github.com/cloudbase/garm/pull/216/commits/0a53b8f6d8635ae1c03d871349566e9df7cb5b5d

cmd/garm/main.go
```diff
@@ -214,13 +214,13 @@ func main() {
...
 	if cfg.Metrics.Enable {
-		slog.InfoContext(ctx, "registering prometheus metrics collectors")
-		if err := metrics.RegisterCollectors(runner); err != nil {
-			log.Fatal(err)
-		}
 		slog.InfoContext(ctx, "setting up metric routes")
 		router = routers.WithMetricsRouter(router, cfg.Metrics.DisableAuth, metricsMiddleware)
+
+		slog.InfoContext(ctx, "start metrics collection")
+		runnerMetrics.CollectObjectMetric(ctx, runner, time.NewTicker(cfg.Metrics.Duration()))
```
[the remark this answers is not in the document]
bavarianbidi: done in https://github.com/cloudbase/garm/pull/216/commits/0a53b8f6d8635ae1c03d871349566e9df7cb5b5d

cmd/garm/main.go
```diff
@@ -214,13 +215,18 @@ func main() {
...
-		if err := metrics.RegisterCollectors(runner); err != nil {
-			log.Fatal(err)
-		}
 		slog.InfoContext(ctx, "setting up metric routes")
 		router = routers.WithMetricsRouter(router, cfg.Metrics.DisableAuth, metricsMiddleware)
+
+		slog.InfoContext(ctx, "register metrics")
+		if err := metrics.RegisterMetrics(); err != nil {
+			log.Fatal(err)
+		}
```
bavarianbidi, line 226: one downside of this implementation (register the metrics here instead of doing it in the `init`):

for any further metric we want to change, we have to do a e.g.
```go
if cfg.Metrics.Enable {
       metrics.InstanceCreatedCount.WithLabelValues(
		bootstrapParams.PoolID, // label: pool_id
		e.cfg.Name,             // label: provider
	).Inc()
}
```

i'm either thinking of registering the metrics, no matter if `cfg.Metrics.Enable` is set or directly drop `cfg.Metrics.Enable` and always run `garm` with `/metrics`.

WDYT @gabriel-samfira ?

gabriel-samfira: Wouldn't we just add it to the `RegisterMetrics()` function and it gets registered here?

gabriel-samfira (approved):

cmd/garm/main.go
gabriel-samfira: In essence, `RegisterMetrics()` would act like the `init()` function, but would be capable of returning an error we can potentially handle. So instead of using the `init()` function, we define our own and call it here to register the metrics.

If metrics are enabled, we register them and start the collector loop.

Issue #214, opened 2024-02-15
Add support for github apps


gabriel-samfira:
We should add support for authenticating garm using [GitHub apps](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app). If nothing else, this should allow a much larger rate limit for github API requests.

PR #228, opened 2024-03-01
Add GitHub App support

gabriel-samfira:
This change adds the ability to use GitHub Apps to authenticate against the GitHub API. This gives us a larger quota for API requests (15k vs 5k for PATs).

Also, each GitHub App has its own quota, whereas PATs share the same user quota.

To use it, define a new credentials config section with the following contents:

```toml
[[github]]
  # This is an arbitrary name
  name = "github_app"
  description = "github app with access to example org/repo"
  # This can be "pat" or "app". Based on this, the HTTP client will either be instantiated
  # using a PAT or an app installation.
  auth_type = "app"
  [github.app]
    # You get this after creating the app. This needs to be changed
    app_id = 1
    private_key_path = "/etc/garm/yourAppName.2024-03-01.private-key.pem"
    # This is the installation ID in your org or repo. This needs to be changed.
    installation_id = 99
```

Fixes: #214


gabriel-samfira:
@Hdom give it a try if you have the time.

PR #232, opened 2024-03-11
Add job info in runner list

### LINKED ISSUE #184: Extend garm-cli runner list
We are using garm on an organization with around 300 repositories and garm is registered on org level. I am wondering if it would be possible to extend `garm-cli runner list -a` (maybe by adding a `-v` option?) to include information on active runners received by the webhook? Something like:
* Repository 
* Name of workflow
* Run ID
* Time when switched to active

So it's easier to see what is currently running?

Optionally it would also be nice to have a command to list the currently queued job to get an easy overview.

gabriel-samfira:
This change adds information about the job a runner is currently handling.

This change also adds a new flag to the CLI `--long | -l` to display this info.

Fixes #184

### DIFF
--- a/cmd/garm-cli/cmd/runner.go
+++ b/cmd/garm-cli/cmd/runner.go
@@ -35,6 +35,7 @@ var (
 	runnerAll            bool
 	forceRemove          bool
 	bypassGHUnauthorized bool
+	long                 bool
 )
 
 // runnerCmd represents the runner command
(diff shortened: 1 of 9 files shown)


gabriel-samfira:
Hi @SystemKeeper !

Is this what you had in mind?

gabriel-samfira:
Looks like this:

image

SystemKeeper:
> Looks like this:
> 
> image

That looks great, exactly what I had in mind 👍

PR #235, opened 2024-03-17
Refactor internal gh client

gabriel-samfira:
This change attempts to refactor some of the code to de-duplicate a bunch of the functions meant to handle github entity types.


gabriel-samfira:
CC @bavarianbidi 

This change also moves metrics counters to the client, ensuring that metrics get counted every time the client is used.

PR #243, opened 2024-04-17
Move github credentials to the database

gabriel-samfira:
This change is part of a larger effort to allow GARM to eventually scale out to multiple instances. The goal is to move as much of the config as possible to the database, allowing a single source of truth for state among multiple instances.

This change also allows us to properly enforce relations between entities (repos, orgs, enterprises), github installations (GHES/github.com) and credentials. This way, we don't end up in an inconsistent state if the credentials are removed from the config, but they're still referenced for an entity in the database.

As part of this change, we aim to maintain as much of the existing user facing API as possible, and also automatically migrate the existing config credentials to the database. The migration is done once, when GARM detects that the needed DB tables are not there. After that, migration is skipped and any credential handling will need to be done using the CLI.

Credentials are scoped to users, and must have a unique name. Different users will be able to create credentials with the same name. The same user will not be able to create multiple credentials with the same name. Although GARM doesn't support multiple users (yet), it may do so in the future. Given the pain of moving the credentials to the DB, this was a decision that was meant to spare us the pain of refactoring later.

Credentials are still referenced by name when assigning them to the entity, but internally we query the DB for a user/credentials combo, given that we have the user ID already stored for the authenticated user, in the context we pass along throughout the code base.

A few more commits are needed to expose an API for handling credentials and to add tests for all the new bits.


gabriel-samfira:
Pfu. The generated code is really bumping up the line count :sweat_smile:

apiserver/routers/routers.go
```diff
@@ -339,16 +339,55 @@ func NewAPIRouter(han *controllers.APIController, authMiddleware, initMiddleware
...
+	// Update Github Endpoint
+	apiRouter.Handle("/github/endpoints/{name}/", http.HandlerFunc(han.UpdateGithubEndpoint)).Methods("PUT", "OPTIONS")
+	apiRouter.Handle("/github/endpoints/{name}", http.HandlerFunc(han.UpdateGithubEndpoint)).Methods("PUT", "OPTIONS")
+
+	////////////////////////
+	// Github credentials //
+	////////////////////////
+	// Legacy credentials path
+	apiRouter.Handle("/credentials/", http.HandlerFunc(han.ListCredentials)).Methods("GET", "OPTIONS")
+	apiRouter.Handle("/credentials", http.HandlerFunc(han.ListCredentials)).Methods("GET", "OPTIONS")
```
bavarianbidi, line 374: not sure if this conflict with the `pathPattern` (defined [here](https://github.com/gabriel-samfira/garm/blob/2a3d524a71adbf5fa5374137fc4073c6d4699e43/client/credentials/credentials_client.go#L167)).

maybe a permanent redirect from `/credentials` to `/github/credentials` might be a better approach:


```suggestion
	// Legacy credentials path
	apiRouter.Handle("/credentials/", http.RedirectHandler("/github/credentials/", http.StatusMovedPermanently).ServeHTTP).Methods("GET", "OPTIONS")
	apiRouter.Handle("/credentials", http.RedirectHandler("/github/credentials", http.StatusMovedPermanently).ServeHTTP).Methods("GET", "OPTIONS")
```

database/sql/sql.go
```diff
@@ -190,6 +195,163 @@ func (s *sqlDatabase) cascadeMigration() error {
 	return nil
 }
 
+func (s *sqlDatabase) ensureGithubEndpoint() error {
+	// Create the default Github endpoint.
+	createEndpointParams := params.CreateGithubEndpointParams{
+		Name:          "github.com",
+		Description:   "The github.com endpoint",
```
bavarianbidi, line 202:
```suggestion
		Name:          "github",
		Description:   "The github endpoint",
```

the `.com` might be a bit confusing in a GitHub-Enterprise context - but no hard meanings on this

apiserver/routers/routers.go
gabriel-samfira: This was left in to allow older clients to still work as expected.  The new client will not use the `/credentials/` API endpoint. Both endpoints call the same handler in the code. I am not sure it's worth adding a redirect, as that generates 2 requests instead of 1, and is only meant to allow older clients to work. This will be removed in future releases of GARM.

It would make sense to use a redirect if you want to notify 3rd party clients or web apps that the endpoint has moved and to update their code. At this point I think that whomever is using GARM, will use the generated go client, or some other client generated from the swagger definitions.

database/sql/sql.go
gabriel-samfira: Before I make changes, I need to add some context around the reasoning behind this.

Right now, the option to use either GitHub or GHES is a part of the credentials we configure in the config. The endpoint is made up of a number of URLs for the base URL )used to compute repo/org/enterprise URLs that are sent to the runners when token config is used), API URL and upload URLs. There is no other info about the endpoint itself.

The assumption was made that there is an extremely low chance that more than one GH installation will use the same exact hostname, but different URL paths. So during the migration of credentials from the config to the DB, it made sense to use the hostname as a `name` for the new DB entry to uniquely identify the endpoint. The name in the DB is also the primary key.

The `github.com` endpoint is the default endpoint that will always exist and which cannot be removed.

I used `github.com` here instead of `github` to be consistent with how a potential GHES endpoint would look like post-migration. In most cases this will look something like `ghes.example.com` or `git.example.com`. But I am too close to the code at this point and may have failed to consider potential confusion. If given the above, you feel that seeing `github.com` instead of `github` as a unique identifier for this endpoint is confusing, I will gladly change it :smile:

gabriel-samfira: I should clarify. When using GHES, a new endpoint will need to be added (if starting from scratch) using:

```bash
garm-cli github endpoint create \
    --base-url https://ghes.example.com \
    --api-base-url https://api.ghes.example.com \
    --upload-url https://upload.ghes.example.com \
    --name example \
    --description "Just an example ghes endpoint" \
    --ca-cert-path $HOME/ca-cert.pem
```

Then you will need to add credentials associated with this new endpoint:

```bash
# for App credential type
garm-cli github credentials add \
    --name deleteme \
    --description "just a test" \
    --endpoint example \
    --auth-type app \
    --app-id 1 \
    --app-installation-id 99 \
    --private-key-path /home/ubuntu/sample.txt
```

```bash
garm-cli github credentials add \
    --name deleteme \
    --description "just a test" \
    --auth-type pat \
    --pat-oauth-token gh_someToken \
    --endpoint example
```

database/sql/sql.go
```diff
@@ -190,6 +195,163 @@ func (s *sqlDatabase) cascadeMigration() error {
...
+			return errors.Wrap(err, "updating repositories")
+		}
+
+		if err := s.conn.Exec("update organizations set credentials_id = ?,endpoint_name = ? where credentials_name = ?", creds.ID, creds.Endpoint, creds.Name).Error; err != nil {
+			return errors.Wrap(err, "updating organizations")
+		}
+
+		if err := s.conn.Exec("update enterprises set credentials_id = ?,endpoint_name = ? where credentials_name = ?", creds.ID, creds.Endpoint, creds.Name).Error; err != nil {
+			return errors.Wrap(err, "updating enterprises")
+		}
```
bavarianbidi, line 350:
```suggestion
		creds, err := s.CreateGithubCredentials(adminCtx, credParams)
		if err != nil {
			return errors.Wrap(err, "creating github credentials")
		}

		if err := s.conn.Exec("update repositories set credentials_id = ?,endpoint_name = ? where credentials_name = ?", creds.ID, creds.Endpoint.Name, creds.Name).Error; err != nil {
			return errors.Wrap(err, "updating repositories")
		}

		if err := s.conn.Exec("update organizations set credentials_id = ?,endpoint_name = ? where credentials_name = ?", creds.ID, creds.Endpoint.Name, creds.Name).Error; err != nil {
			return errors.Wrap(err, "updating organizations")
		}

		if err := s.conn.Exec("update enterprises set credentials_id = ?,endpoint_name = ? where credentials_name = ?", creds.ID, creds.Endpoint.Name, creds.Name).Error; err != nil {
			return errors.Wrap(err, "updating enterprises")
		}
```

`creds.Endpoint` is of type `GithubEndpoint` ([xref](https://github.com/gabriel-samfira/garm/blob/8726cb994eeeee709a98ae49f2d793f35646d6a1/params/params.go#L825-L834)) and can be converted.

in my test, garm-server crashed with the following output:

```
sql: converting argument $2 type: unsupported type params.GithubEndpoint, a struct\nupdating repositories\ngithub.com/cloudbase/garm/database/sql.(*sqlDatabase).migrateCredentialsToDB\n\t/go/garm_repo/database/sql/sql.go:341\ngithub.com/cloudbase/garm/database/sql.(*sqlDatabase).migrateDB\n\t/go/garm_repo/database/sql/sql.go:428\ngithub.com/cloudbase/garm/database/sql.NewSQLDatabase\n\t/go/garm_repo/database/sql/sql.go:77\ngithub.com/cloudbase/garm/database.NewDatabase\n\t/go/garm_repo/database/database.go:30\nmain.main\n\t/go/garm_repo/cmd/garm/main.go:175\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:267\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1650"
time=2024-05-07T09:57:43.562Z level=ERROR msg="rolling back github github endpoints table" error="sql: converting argument $2 type: unsupported type params.GithubEndpoint, a struct\nupdating repositories\ngithub.com/cloudbase/garm/database/sql.(*sqlDatabase).migrateCredentialsToDB\n\t/go/garm_repo/database/sql/sql.go:341\ngithub.com/cloudbase/garm/database/sql.(*sqlDatabase).migrateDB\n\t/go/garm_repo/database/sql/sql.go:428\ngithub.com/cloudbase/garm/database/sql.NewSQLDatabase\n\t/go/garm_repo/database/sql/sql.go:77\ngithub.com/cloudbase/garm/database.NewDatabase\n\t/go/garm_repo/database/database.go:30\nmain.main\n\t/go/garm_repo/cmd/garm/main.go:175\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:267\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1650"
time=2024-05-07T09:57:43.562Z level=INFO msg="migrating database: migrating credentials: updating repositories: sql: converting argument $2 type: unsupported type params.GithubEndpoint, a struct"
```

gabriel-samfira: Added a DB migration test.

database/sql/sql.go
bavarianbidi: ignore my comment - just stumpled over `github.com` as string and was confused. after a first test in our setup, i've seen that this `github.com` string appears in the default endpoint configuration:

The github Endpoint configuration is stored in a dedicated table: (`select * from github_endpoints;`)

| name|created_at|updated_at|deleted_at|description|api_base_url|upload_base_url|base_url|ca_cert_bundle
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
github.com|2024-05-07 11:34:12.171717827+00:00|2024-05-07 11:34:12.171717827+00:00||The github.com endpoint|https://api.github.com/|https://uploads.github.com/|https://github.com|
git.company.com |2024-05-07 11:34:12.172863756+00:00|2024-05-07 11:34:12.172863756+00:00||Endpoint for git.company.com |https://git.company.com |https://git.company.com |https://git.company.com |

apiserver/routers/routers.go
bavarianbidi: valid point - using the correct cli will solve this issue - i'm fine with that

Issue #248, opened 2024-05-21
Pools shouldn't contain the default labels `self-hosted`, `$OS_ARCH` and `$OS_TYPE`


bavarianbidi:
To get action runners without default-labels, the following things has to be done

### ToDos
- [x] garm-provider-common has to make use of the `--no-default-labels` flag :heavy_check_mark: 
- [x] garm-provider-common must be released :hourglass_flowing_sand: 
- [x] providers make use of the new garm-provider-common version
     - [x] openstack
     - [x] azure
     - [x] gcp
     - [x] oci
     - [x] aws
     - [x] lxd
     - [x] incus
     - [x] equinix
- [x] garm-pools shouldn't do any default-label magic any more. Implementation is ongoing [here](https://github.com/cloudbase/garm/compare/main...mercedes-benz:garm:new_runners_without_default_labels?expand=1) 
- [x] describe new behavior

gabriel-samfira:
also, any other external providers that use `garm-provider-common` will need to update before being used with the next version of GARM. Or any other provider that does not use the common package, but sets up the runner using a registration token.

gabriel-samfira:
I accidentally created an issue by clicking a circle next to the text. GutHub AI magic I guess 😅

PR #249, opened 2024-05-21
New runners without default labels

bavarianbidi:
towards https://github.com/cloudbase/garm/issues/248

### DIFF
--- a/doc/labels.md
+++ b/doc/labels.md
@@ -10,6 +10,6 @@ These are:
 [ 'self-hosted', '$OS_TYPE', '$OS_ARCH' ]
 ```
 
-With Version **TBD** of `garm-provider-common`, the runner script will register themselves with a new command line flag, called `--no-default-labels`. If this flag is set, the runner will not append any default label.
+With Version `v0.1.2` of `garm-provider-common`, the runner script will register themselves with a new command line flag, called `--no-default-labels`. If this flag is set, the runner will not append any default label.
 
 As all labels can be defined in the pool specification, it's still possible to add the default labels manually.
--- a/doc/labels.md
+++ b/doc/labels.md
@@ -1,0 +1,15 @@
+# Labels and Tags
+
+Github Runners can be tagged with labels. These labels can be used to restrict the jobs that can run on a runner. For example, you can have a runner with the label `linux` and another with the label `windows`. You can then restrict a job to run only on a runner with the label `linux`.
+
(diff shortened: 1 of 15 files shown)


doc/labels.md
```diff
@@ -0,0 +1,15 @@
...
+
+Whenever a new runner register themselves on Github, the runner knows its own labels as the labels are defined in the pool specification as tags.
+
+Beside the custom labels, Github also has some predefined labels that are appended by the runner script per default.
+These are: 
+```yaml
+[ 'self-hosted', '$OS_TYPE', '$OS_ARCH' ]
+```
+
+With Version **TBD** of `garm-provider-common`, the runner script will register themselves with a new command line flag, called `--no-default-labels`. If this flag is set, the runner will not append any default label.
```
bavarianbidi, line 13: @gabriel-samfira i would like to mention the version of `garm-provider-commons` here to make it a bit more clearer since when the runner starts with `--no-default-labels`

runner/pools.go
```diff
@@ -99,12 +99,8 @@ func (r *Runner) UpdatePoolByID(ctx context.Context, poolID string, param params
 		return params.Pool{}, runnerErrors.NewBadRequestError("min_idle_runners cannot be larger than max_runners")
 	}
 
-	if param.Tags != nil && len(param.Tags) > 0 {
-		newTags, err := r.processTags(string(pool.OSArch), pool.OSType, param.Tags)
-		if err != nil {
-			return params.Pool{}, errors.Wrap(err, "processing tags")
-		}
-		param.Tags = newTags
+	if len(param.Tags) == 0 {
```
gabriel-samfira, line 102: It's not mandatory to send labels as part of the update request. Sending labels is only mandatory during create operations.

Given that we no longer have to append the default labels to supplied labels (if any were supplied), we can just remove any check made against `param.Tags`

doc/labels.md
gabriel-samfira: yup. Will cut a new release of common and update here. Hopefully early tomorrow morning.

gabriel-samfira (changes requested):
Thanks for this! I've been meaning to make this change ever since we added the `--no-default-labels` in the github actions runner :sweat_smile:

doc/labels.md
gabriel-samfira: https://github.com/cloudbase/garm-provider-common/releases/tag/v0.1.2

runner/pools.go
bavarianbidi: done

gabriel-samfira (approved):

PR #255, opened 2024-06-05
fix: remove unnecessary github api call

bavarianbidi:
There are only a few cases, where we get a job information from github where the runner name is not set.

For all this cases we do not need to check github API at all because these jobs are never ever get scheduled to a runner:

job.Action is:

* queued: a queued job is just queued and not scheduled to a runner so we do not get a runner name from the GH API
* completed: when conclusion=cancelled|failure github never scheduled the job to a runner and with that we do not get a runner name from the GH API

### DIFF
--- a/runner/pool/pool.go
+++ b/runner/pool/pool.go
@@ -35,7 +35,6 @@ import (
 	"github.com/cloudbase/garm-provider-common/util"
 	"github.com/cloudbase/garm/auth"
 	dbCommon "github.com/cloudbase/garm/database/common"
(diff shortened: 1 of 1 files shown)


gabriel-samfira (approved):

gabriel-samfira:
Could you add a check here:

https://github.com/cloudbase/garm/blob/02ff74bfe89e85bb0d9a689a99f2ea3169d113d8/runner/pool/pool.go#L212

For the runner name. And just log that there was no runner assigned and just return nil.

Saves us some unnecessary db queries and log spam

PR #263, opened 2024-06-14
Add database watcher

gabriel-samfira:
Adds a database watcher framework that will allow us to react to changes made to entities in the database.


gabriel-samfira:
Merging this as is. We can iterate.

PR #275, opened 2024-07-04
Add event stream

gabriel-samfira:
This change adds a new websocket endpoint for database events. The events endpoint allows clients to stream events as they happen in GARM. Events are defined as a structure containning the event type (create, update, delete), the database entity involved (instances, pools, repos, etc) and the payload consisting of the object involved in the event. The payload translates to the types normally returned by the API and can be deserialized as one of the types present in the params package.

The events endpoint is a websocket endpoint and it accepts filters as a json encoded `Options{}` type [defined here](https://github.com/gabriel-samfira/garm/blob/06e2e33e1f89379c7dbdbc9b6ea076a844e52f67/apiserver/events/params.go#L32-L35). This json is send over the websocket connection. The filters allows the user to specify which entities are of interest, and which operations should be returned. For example, you may be interested in changes made to pools or runners, in which case you could create a filter that only returns update operations for pools. Or update and delete operations.

The filters can be defined as:
```json
{
  "filters": [
    {
      "entity_type": "instance",
      "operations": ["update", "delete"]
    },
    {
      "entity_type": "pool"
    },
  ],
  "send_everything": false
}
```

This would return only update and delete events for instances and all events for pools. Alternatively you can ask GARM to send you everything:

```json
{
  "send_everything": true
}
```

If no ofilters are set, all events are ignored and nothing is sent over the connection.

To help view these events, a new command was added to the `garm-cli` command:

```bash
garm-cli debug-events \
    --filters='{"send_everything": false, "filters": [{"entity_type": "instance"}, {"entity_type": "pool"}]}'
```

Which will return events for pools and instances. Here is an example of the command in action:

[asciicast](https://asciinema.org/a/crIRgWPq0Mj6tFbA6gRs6T7Rd)

More complex filters may be added in the future.

A few notes on authentication. To authenticate against the websocket endpoint, you'll still use the JWT bearer token you normally use. The websocket handler will then start a timer that will expire at the same time as your token. When that happens, your connection will be terminated.

Also, if the user is updated and either is disabled, the password changes or is removed, the connection will be terminated.


gabriel-samfira:
merging this as is. We can iterate later.

PR #278, opened 2024-07-22
Add provider interface versioning

fabi200123:
This PR adds provider interface versioning:
- version **v0.1.0** will be the old implementation of external provider interface
- version **v0.1.1** will add in each command's Environment the `Interface Version`, `Pool Info` and `Controller Info`
- **default version** will remain **v0.1.0**

> **NOTE** This PR is using garm-provider-common version defined in [this PR](https://github.com/cloudbase/garm-provider-common/pull/30) and must be updated accordingly after merging the PR from garm-provider-common.


runner/providers/external/external.go
```diff
@@ -2,334 +2,22 @@ package external
...
-// been updated to support JIT configuration.
-func (e *external) DisableJITConfig() bool {
-	if e.cfg == nil {
-		return false
+	switch cfg.External.InterfaceVersion {
+	case "v0.1.0":
+		return v010.NewProvider(ctx, cfg, controllerID)
+	case "v0.1.1":
+		return v011.NewProvider(ctx, cfg, controllerID)
+	default:
```
gabriel-samfira, line 19: on default we should probably err out. We should default empty string to `0.1.0` though.

runner/providers/external/external.go
```diff
@@ -2,334 +2,22 @@ package external
...
-}
-
-// DisableJITConfig tells us if the provider explicitly disables JIT configuration and
-// forces runner registration tokens to be used. This may happen if a provider has not yet
-// been updated to support JIT configuration.
-func (e *external) DisableJITConfig() bool {
-	if e.cfg == nil {
-		return false
+	switch cfg.External.InterfaceVersion {
+	case "v0.1.0":
```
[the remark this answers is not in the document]
gabriel-samfira: Also, this can probably be:

```go
case "v0.1.0", "":
```

Which will mean an explicit `0.1.0` or if not set, default to `0.1.0`

runner/providers/v0.1.1/external.go
```diff
@@ -0,0 +1,387 @@
...
+	if err != nil {
+		return commonParams.ProviderInstance{}, errors.Wrap(err, "serializing extraspecs")
+	}
+	asEnv := []string{
+		fmt.Sprintf("GARM_COMMAND=%s", execution.CreateInstanceCommand),
+		fmt.Sprintf("GARM_CONTROLLER_ID=%s", e.controllerID),
+		fmt.Sprintf("GARM_POOL_ID=%s", bootstrapParams.PoolID),
+		fmt.Sprintf("GARM_PROVIDER_CONFIG_FILE=%s", e.cfg.External.ConfigFile),
+		fmt.Sprintf("GARM_POOL_IMAGE=%s", createInstanceParams.CreateInstanceV011.PoolInfo.Image),
+		fmt.Sprintf("GARM_POOL_EXTRASPECS=%s", string(extraspecsValue)),
```
gabriel-samfira, line 86: a json may have all sorts of characters that may break shells. Have this be a base64 encoded string. That will ensure no weirdness will happen.

runner/providers/v0.1.1/external.go
```diff
@@ -0,0 +1,387 @@
...
+	extraspecsValue, err := json.Marshal(extraspecs)
+	if err != nil {
+		return commonParams.ProviderInstance{}, errors.Wrap(err, "serializing extraspecs")
+	}
+	asEnv := []string{
+		fmt.Sprintf("GARM_COMMAND=%s", execution.CreateInstanceCommand),
+		fmt.Sprintf("GARM_CONTROLLER_ID=%s", e.controllerID),
+		fmt.Sprintf("GARM_POOL_ID=%s", bootstrapParams.PoolID),
+		fmt.Sprintf("GARM_PROVIDER_CONFIG_FILE=%s", e.cfg.External.ConfigFile),
+		fmt.Sprintf("GARM_POOL_IMAGE=%s", createInstanceParams.CreateInstanceV011.PoolInfo.Image),
```
gabriel-samfira, line 85: I think this is already set in the bootstrap we send to stdin.

runner/providers/v0.1.1/util.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	commonParams "github.com/cloudbase/garm-provider-common/params"
+)
+
+// IsProviderValidStatus checks if the given status is valid for the provider.
+// A provider should only return a status indicating that the instance is in a
+// lifecycle state that it can influence. The sole purpose of a provider is to
+// manage the lifecycle of an instance. Statuses that indicate an instance should
+// be created or removed, will be set by the controller.
+func IsValidProviderStatus(status commonParams.InstanceStatus) bool {
```
gabriel-samfira, line 12: This can probably live in a common package considering this validates what GARM expects from any provider (external or otherwise)

gabriel-samfira (changes requested):
First pass. Will do another tomorrow.

runner/providers/v0.1.1/external.go
fabi200123: Indeed, removed it for CreateInstance. For the moment, changed it to `CreateInstance(ctx context.Context, bootstrapParams commonParams.BootstrapInstance, _ common.CreateInstanceParams)` since we don't use (for now) anything from there

runner/providers/v0.1.1/util.go
fabi200123: Created a new `util` package in `runner/providers/util` for this common part

runner/providers/v0.1.1/external.go
```diff
@@ -0,0 +1,394 @@
...
+
+	return nil
+}
+
+// CreateInstance creates a new compute instance in the provider.
+func (e *external) CreateInstance(ctx context.Context, bootstrapParams commonParams.BootstrapInstance, _ common.CreateInstanceParams) (commonParams.ProviderInstance, error) {
+	asEnv := []string{
+		fmt.Sprintf("GARM_COMMAND=%s", commonExecution.CreateInstanceCommand),
+		fmt.Sprintf("GARM_CONTROLLER_ID=%s", e.controllerID),
+		fmt.Sprintf("GARM_POOL_ID=%s", bootstrapParams.PoolID),
```
gabriel-samfira, line 80: to be consistent with the rest of the functions, I think we can have the `GARM_POOL_EXTRASPECS` variable set here as well. I know the `CreateInstance` command will look for extra specs in the bootstrap instance stdin body, but we should be consistent. The fewer differences in terms of env variables, the better.

runner/pool/pool.go
```diff
@@ -589,7 +589,15 @@ func (r *basePoolManager) cleanupOrphanedGithubRunners(runners []*github.Runner)
 			slog.DebugContext(
 				r.ctx, "updating instances cache for pool",
 				"pool_id", pool.ID)
-			poolInstances, err = provider.ListInstances(r.ctx, pool.ID)
+			listInstancesParams := common.ListInstancesParams{
+				ListInstancesV011: common.ListInstancesV011Params{
+					ProviderBaseParams: common.ProviderBaseParams{
```
gabriel-samfira, line 594: At this point, the base params will always contain the same thing. Please create a function in the `basePoolManager{}` that returns the `ProviderBaseParams{}` struct. And you can just call it.

The function should be protected by `r.mux.Lock()` as the pool info can be changed and updated by the DB watcher.

gabriel-samfira: We might also want to switch on the interface version in the future and return the appropriate `ListInstancesParams{}`

gabriel-samfira (changes requested):
Round 1 review. Will do another review soon.

runner/pool/pool.go
```diff
@@ -654,7 +669,15 @@ func (r *basePoolManager) cleanupOrphanedGithubRunners(runners []*github.Runner)
 				r.ctx, "instance was found in stopped state; starting",
 				"runner_name", dbInstance.Name)
 
-			if err := provider.Start(r.ctx, dbInstance.ProviderID); err != nil {
+			startParams := common.StartParams{
+				StartV011: common.StartV011Params{
+					ProviderBaseParams: common.ProviderBaseParams{
+						PoolInfo:       pool,
```
gabriel-samfira, line 675: Perhaps `r.getProviderBaseParams(pool)`. Here and everywhere else?

gabriel-samfira (approved):

PR #302, opened 2024-10-07
Add Using Cached Runner documentation

fabi200123:
- Adds documentation for using cached GitHub action runners for Windows and Linux

Not to be merged before the following PRs are merged:

- [x] [garm-provider-common - #36](https://github.com/cloudbase/garm-provider-common/pull/36)
- [x] [garm-provider-common - #37](https://github.com/cloudbase/garm-provider-common/pull/37)

### DIFF
--- a/doc/using_cached_runners.md
+++ b/doc/using_cached_runners.md
@@ -1,0 +1,52 @@
+# Using Cached Runners
+
+## GitHub Action Runners and GARM
+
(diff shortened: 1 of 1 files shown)


gabriel-samfira:
I think we can refactor the performance_considerations.md doc and combine it with this. Or move the LXD specific docs to the LXD/inclus repos and replace the performance considerations with this. What do you think?

fabi200123:
I think that a better approach here would be to move the LXD/Incus specific stuff to its provider repo. Then perhaps create a similar file for each provider repository. I think that would answer most people's questions regarding cached runners and perhaps even let them know they can use the cached runners

PR #314, opened 2024-11-22
Improve error messages in garm log

maigl:
finding errors in logs is very important and there are some logs that are `errors` but seem to be rather normal:

- workflow events without labels -> this happens very often (at least in our GHE instance), this might be a bug in github code .. but garm probably has to accept this and don't consider this an error
- workflows that have runner_names which are not `our` runners. This always happens in a scenario where our garm runners are not the only runner setup (and we still get the workflow_job events, e.g. enterprise wide hooks)

### DIFF
--- a/database/sql/jobs.go
+++ b/database/sql/jobs.go
@@ -3,6 +3,7 @@ package sql
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log/slog"
 
 	"github.com/google/uuid"
@@ -84,7 +85,8 @@ func (s *sqlDatabase) paramsJobToWorkflowJob(ctx context.Context, job params.Job
 	if job.RunnerName != "" {
 		instance, err := s.getInstanceByName(s.ctx, job.RunnerName)
 		if err != nil {
-			slog.With(slog.Any("error", err)).ErrorContext(ctx, "failed to get instance by name")
+			// This usually is very normal as not all jobs run on our runners.
+			slog.DebugContext(ctx, fmt.Sprintf("failed to get instance by name: %s", job.RunnerName))
 		} else {
 			workflofJob.InstanceID = &instance.ID
 		}
@@ -244,7 +246,8 @@ func (s *sqlDatabase) CreateOrUpdateJob(ctx context.Context, job params.Job) (pa
 			if err == nil {
 				workflowJob.InstanceID = &instance.ID
 			} else {
-				slog.With(slog.Any("error", err)).ErrorContext(ctx, "failed to get instance by name")
+				// This usually is very normal as not all jobs run on our runners.
+				slog.DebugContext(ctx, fmt.Sprintf("failed to get instance by name: %s", job.RunnerName))
 			}
 		}
 
--- a/database/sql/jobs.go
+++ b/database/sql/jobs.go
@@ -3,7 +3,6 @@ package sql
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"log/slog"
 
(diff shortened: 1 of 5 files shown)


database/sql/jobs.go
```diff
@@ -84,7 +85,8 @@ func (s *sqlDatabase) paramsJobToWorkflowJob(ctx context.Context, job params.Job
 	if job.RunnerName != "" {
 		instance, err := s.getInstanceByName(s.ctx, job.RunnerName)
 		if err != nil {
-			slog.With(slog.Any("error", err)).ErrorContext(ctx, "failed to get instance by name")
+			// This usually is very normal as not all jobs run on our runners.
+			slog.DebugContext(ctx, fmt.Sprintf("failed to get instance by name: %s", job.RunnerName))
```
gabriel-samfira, line 89: Let's try something like this:

```suggestion
			slog.With(slog.Any("error", err)).DebugContext(ctx, "failed to get instance by name", "instance_name", job.RunnerName)
```

slog allows us to have structured logging, where we can add keys and values to those keys. No need to use `fmt` to format the messages.

in the suggestion above, if you export the log in json format, you will have the message `failed to get instance by name` and 2 keys: `error: "error message goes here"` and `instance_name: "garm-XXXXX"`. When printing as text, they'll show up as `error=message` and `instance_name=garm-XXXX`.

database/sql/jobs.go
```diff
@@ -244,7 +246,8 @@ func (s *sqlDatabase) CreateOrUpdateJob(ctx context.Context, job params.Job) (pa
 			if err == nil {
 				workflowJob.InstanceID = &instance.ID
 			} else {
-				slog.With(slog.Any("error", err)).ErrorContext(ctx, "failed to get instance by name")
+				// This usually is very normal as not all jobs run on our runners.
+				slog.DebugContext(ctx, fmt.Sprintf("failed to get instance by name: %s", job.RunnerName))
```
gabriel-samfira, line 250: same as above. The level change to `DebugContext` is fine.

gabriel-samfira (changes requested):
Thanks! I should have added more context to those messages.

database/sql/jobs.go
maigl: .. okay, actually I was aiming to remove the word 'error' from those log lines, I usually grep for error and this is not really an error ..even if there is a error-struct  thrown somewhere. Of course, I can change my grepping .. but do you see it's valuable to see the error message ?

gabriel-samfira: In this context, I think we can just remove it. It's just spam.

gabriel-samfira (approved):

gabriel-samfira:
I think we need to bump `golangci-lint` to >= 1.60.1. Details here: https://github.com/golangci/golangci-lint/issues/4662#issuecomment-2323425663

maigl:
I added a specific `golangci-lint` version in the Makefile .. now we're left with a bunch of G115: integer overflow conversion findings .. I think they are okayish .. how to deal with them, exclude G115?

gabriel-samfira:
Yeah. Let's exclude them for now. I don't think anyone will try to set 2^32 MaxRunners. Anytime soon. We may change the types later to avoid the need to convert, or add some overflow checks. But for now, let's jut exclude them.

maigl:
> Yeah. Let's exclude them for now. I don't think anyone will try to set 2^32 MaxRunners. Anytime soon. We may change the types later to avoid the need to convert, or add some overflow checks. But for now, let's jut exclude them.

done

PR #315, opened 2024-11-26
fix: remove duplication in Makefile

maigl:
I created some duplication in Makefile. I'm sorry ..


gabriel-samfira:
That's okay @maigl . May this be the worst thing to happen today and we can call it a great day.

PR #328, opened 2025-01-28
Add knob to tweak _busy_timeout

gabriel-samfira:
In SQLite3 we may need to set a busy_timeout in the case of instances with high load. This change adds a knob that allows users to set a timeout if a database is locked for writing by another routine.

for details about the SQLIte3 option, see: https://www.sqlite.org/c3ref/busy_timeout.html


gabriel-samfira:
CC @maigl

PR #329, opened 2025-01-28
Add a backoff mechanism when deleting runners

gabriel-samfira:
This change adds a backoff mechanism when deleting github runners. If the delete operation fails, we record the event and retry with a geometric progression of 1.5 starting from 5 seconds, which is the pool consolidation timeout.


gabriel-samfira:
CC @maigl

Issue #323, opened 2024-12-24
Gitea support


hulto:
Hey!
I just started playing with Garm and I'm really enjoying it!
I'd like to use it with my gitea server as well as some GitHub repos.

Are there plans to add support for Gitea? If not is that something I can contribute?

gabriel-samfira:
Hi @hulto!

As long as Gitea/Forgejo supports:
* ephemeral runners
* registration tokens/JIT runners
* can send webhooks with job runs so we can schedule runners to a pool
* Has a way to allow us to list compatible agent downloads for linux/windows/mac/etc

It should be possible. However, I think there is an inherent danger in trying to support 2 different forges in the same project, as they may deviate significantly in the future and it may become cumbersome or downright impossible to consolidate. It may be wise to attempt to rearchitect parts of GARM and split off the bits that deal with the interaction between GARM and the forge. Similar to how providers work. But that would be a huge undertaking.

In the short term, if gitea has the above mentioned features, an attempt could be made to support it, if you have the time. Up to this point the focus has been to support GitHub and GHES. So the code mostly revolves around that assumption. But I think some thin abstractions could make gitea fit.

My experience with gitea is limited, so I am not sure how much I could help, but if I can, I will. If you want to tackle this, I suggest creating a proof of concept, hacked with an axe, just to see if it's feasible. Something that won't take months of your time, just to find that we can't merge it.

If having the two forges coexist does not seem possible, another option would be to fork GARM, yank out github and replace it with gitea. The we can merge back if/when the architecture of GARM permits pluggable forges, as we could then separately maintain plugins for gitea, github, etc.

So, plenty of options 😊.

gabriel-samfira:
Hi @ChristopherHX 

Adding Gitea will be a bit more involved. There are a few assumptions in GARM that are related to GitHub. Each entity (Repo, Org, Enterprise) has a "pool manager" (needs to be reworked). That pool manager makes sure that the runners in each pool are properly created, deleted, etc. To ensure that GARM itselt has valid credentials to manage resources inside an entity, we periodically fetch the tools. We need them anyway, and in some cases, tools have a bearer token we need to ass to be able to download them (GHES). So checking for tools has a dual purpose:

* Validate credentials
* keep the tools object in cache for the entity

For Gitea we don't have enterprises, which is fine. But if Gitea doesn't have "organizations", only repos and global, we need to create a new entity type.

[...]

PR #393, opened 2025-05-14
Add gitea

gabriel-samfira:
This branch adds initial Gitea support. What works:

* Adding gitea endpoint
* adding gitea credentials
* adding repositories
* adding orgs
* Webhooks properly record jobs
* Runners are managed properly
* Installing webhooks
* PAT auth

What doesn't work:

* ~~Orgs (API endpoints and CLI need to be added, the rest is there)~~
* ~~automatica webhook install. Gitea requires webhook type. We need to use the official SDK.~~
* System and user level entities. Gitea doesn't have "enterprises".
* Application auth.

~~Right now, we use the go-github client to interact with Gitea, but that has limitations. Before merging, we'll have to switch to the official SDK.~~ For the initial support, we might stick with go-github.

TODO: ~~fix existing tests~~, write new ones

Steps for testing this feature out:

## Create a gitea endpoint

Gitea being self hosted, you will need to add a new endpoint, the same way you would do for GHES. This only needs to be done once for a Gitea instance.

For the purpose of this example, we'll assume that the endpoint is reachable at `http://10.0.9.5/` (I'm just too lazy to adapt my bash history).

```bash
garm-cli gitea endpoint create \
    --api-base-url http://10.0.9.5/ \
    --base-url http://10.0.9.5/ \
    --description "My first Gitea endpoint" \
    --name local-gitea
```

## Add gitea credentials

This branch supports PATs with Gitea. Apps may be added in the future. You will need to generate a PAT with read/write access to repositories and orgs. Then we can add the credentials to GARM:

```bash
garm-cli gitea credentials add \
    --endpoint local-gitea \
    --auth-type pat \
    --pat-oauth-token yourSuperSecretPAT \
    --name gsamfira-token \
    --description "Gitea token"
```

You now have a Gitea endpoint and credentials tied to it.

## Create a repository

Repositories and Orgs are common for both GitHub and Gitea. We use the same database model for both with slight changes. Future entities (System and User) will be added as distinct models, available only for gitea, the same way enterprises are exclusive to GitHub.

But I digress. Adding the repo:

```bash
garm-cli repo add \
    --credentials gsamfira-token \
    --name testrepo \
    --owner gsamfira \
    --random-webhook-secret \
    --install-webhook \
    --forge-type gitea
```

Make a note of the repo UUID. You will need it when defining the pool.

~~Important note: At this stage, GARM can't yet install webhooks in Gitea. It can remove it (which happens by default when removing a repo from GARM), but creating it diverges from how github works. We will use the official SDK before merging this change, so we can expect webhook management to work. But for now, instead of `--random-webhook-secret`, you will have to explicitly set a `--webhook-secret` and make sure you use the same secret when creating the webhook in the repo or org.~~

Installing webhooks now works.

## Adding a pool

For testing I used the `LXD` provider, but you can use any of the existing supported providers, which offer a way to override the runner install template.

For Gitea to work, the provider needs to know how to install the act runner. This is done by an installation script which is generated from a template. We can override the default template in most providers using the `runner_install_template` field in `extra_specs`. You can download a sample extra specs [from here](https://gist.githubusercontent.com/gabriel-samfira/d132169ec41d990bbe17e6097af94c4c/raw/67d226d9115eca5e10b69eac5ecb04be91a48991/gitea-extra-specs.json). Feel free to customize.

Download the extra specs file:

```bash
curl -s -L https://gist.githubusercontent.com/gabriel-samfira/d132169ec41d990bbe17e6097af94c4c/raw/67d226d9115eca5e10b69eac5ecb04be91a48991/gitea-extra-specs.json -o ~/garm-extra-specs.json
```

Create the pool:

```bash
garm-cli pool add \
    --repo THE_UUID_OF_THE_REPO_YOU_CREATED_ABOBVE \
    --provider-name arrakis \
    --image ubuntu:24.04 \
    --tags ubuntu-latest \
    --flavor default \
    --extra-specs-file $HOME/garm-extra-specs.json \
    --enabled=true \
    --min-idle-runners=1
```

You should soon see an idle runner being spun up in the provider and associated to your gitea instance. You should now be able to create a new workflow and target the runner with the `ubuntu-latest` label. You can add multiple labels, set a different image or `flavor`. 

GARM will automatically spin up runners within the constraints set by the pool (min/max runners).

Fixes: #323


gabriel-samfira:
> This client is very very extensible,

I tend to prefer to use official clients if available. It saves us from having to maintain that code as well. 

> Your existing e2e test assumes an existing forge instance e.g. github.

Yep. It's too early to think about e2e for gitea. But the easiest way would be to just set up gitea on the runner that picks up the job. Gitea even has a snap from what I see so it should be really easy to set up on any VM. Even creating a custom image with gitea pre-installed and just spinning that up via GARM itself would do.

So we will see. First, I need to add `orgs` and perhaps `system` as an entity, then unit tests, then integration tests.

gabriel-samfira:
@ChristopherHX 

Interesting. Apparently gitea doesn't properly validate the webhook config. I can create a new webhook via the API, without an actual "target URL". The UI validates correctly. The API does not. Notice the missing URL bellow:

image

And here you can see I'm editing hook with ID 14 (visible in the URL bar), but the target URL is empty:

image


Using this body:

```go
createOpts := &createGiteaHookOptions{
		Type:         "gitea",
		Events:       hook.Events,
		Active:       hook.GetActive(),
		BranchFilter: "*",
		Config: map[string]string{
			"content_type": hook.GetConfig().GetContentType(),
			"url":          "",
			"http_method":  "post",
		},
	}
```

The struct is defined as:

```go
type createGiteaHookOptions struct {
	Type                string            `json:"type"`
	Config              map[string]string `json:"config"`
	Events              []string          `json:"events"`
	BranchFilter        string            `json:"branch_filter"`
	Active              bool              `json:"active"`
	AuthorizationHeader string            `json:"authorization_header"`
}
```

gabriel-samfira:
added a big report here:

  * https://github.com/go-gitea/gitea/issues/34491

lunny:
> * Webhooks - API does not validate target URL go-gitea/gitea#34491

https://github.com/go-gitea/gitea/pull/34492

gabriel-samfira:
Added orgs. I think that's a sufficient target for initial support. Most people will probably need repo and org level runners.

Will focus on tests from now on. While this branch may work, for anyone that wants to test it, expect bugs.

gabriel-samfira:
Merging this as is. New PRs will be made with integration tests and more unit tests. The pool manager will also need to be migrated to its own worker similar to scalesets and the rest before we can make a new release.

PR #416, opened 2025-05-29
Create Repo / Org make --forge-type optional

ChristopherHX:
* This makes gitea garm usage unnessary complex

### DIFF
--- a/cmd/garm-cli/cmd/organization.go
+++ b/cmd/garm-cli/cmd/organization.go
@@ -308,7 +308,7 @@ func init() {
 	orgAddCmd.Flags().StringVar(&orgName, "name", "", "The name of the organization")
 	orgAddCmd.Flags().StringVar(&poolBalancerType, "pool-balancer-type", string(params.PoolBalancerTypeRoundRobin), "The balancing strategy to use when creating runners in pools matching requested labels.")
 	orgAddCmd.Flags().StringVar(&orgWebhookSecret, "webhook-secret", "", "The webhook secret for this organization")
-	orgAddCmd.Flags().StringVar(&forgeType, "forge-type", string(params.GithubEndpointType), "The forge type of the organization. Supported values: github, gitea.")
+	orgAddCmd.Flags().StringVar(&forgeType, "forge-type", "", "The forge type of the organization. Supported values: github, gitea.")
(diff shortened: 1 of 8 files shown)


cmd/garm-cli/cmd/common.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/cloudbase/garm/client/credentials"
+	"github.com/cloudbase/garm/params"
+)
+
+func queryForgeType(credentialName string) error {
+	if forgeType == "" {
+		// TODO do this on server?, where we can query by Name
+		list, err := apiCli.Credentials.ListGiteaCredentials(credentials.NewListGiteaCredentialsParams(), authToken)
```
gabriel-samfira, line 11: Gitea and Github credentials each have their own model, so for consistency, we would have to query both and see which one has the creds with that name. If both have identically named creds, we would need to err out and ask for the --forge-type.

I think this would be more efficient if implemented server side: https://github.com/cloudbase/garm/blob/18ef27bb9138f49f861171e6db14ca13c85910c1/runner/organizations.go#L32-L51

It would save the 2 extra API calls. If we omit the forge type, we could use the default branch of the switch to query both models. Right now, `param.GetForgeType()` defaults to github for backwards compatibility.

If we do this server side, then the client can just drop the default forge type and leave the field empty.

gabriel-samfira (approved):
As I wanted to submit a comment, you pushed a commit that solved that comment and GH erred out because of the outdated commit I was commenting on :smile: .

This LGTM. If tests pass, we can merge.

gabriel-samfira:
To fix the lint error, you can probably run:

```bash
golangci-lint run --build-tags=testing --fix
```

gabriel-samfira:
> more like, due to linuxism

probably due to the shell used. Changing:

https://github.com/cloudbase/garm/blob/18ef27bb9138f49f861171e6db14ca13c85910c1/Makefile#L70

to 

```Makefile
@cp -R ${ROOTDIR} ${TMPDIR}/.
```

will make it work on your system as well.

gabriel-samfira:
Thanks!

gabriel-samfira:
> Actually not, this was for example macOS and the error is of a different kind.

if you have a bit of RAM to spare, you could probably use multipass on Mac to get a quick ubuntu VM up and running locally.

PR #421, opened 2025-05-29
Allow owner/repo as repoid if unique

ChristopherHX:
This one is so much easier on the client, the problem here is owner/repo contains a slash.

The slash is escaped, gorilla and several reverse proxies will merge %2f as `/` and change the url target.

Disabling this somehow looks awful, and may not work correctly depending what is between garm and the cli client

- UseEncodedPath disables path merging of %2f escape for /
- unescapeVars reverts the missing decoding after the api is called
- GetRepository allows to be called with an empty endpointName, returns an error if it is not unique 

_This does not feel right, unless moving repoid out of the PATH_

### DIFF
--- a/apiserver/routers/routers.go
+++ b/apiserver/routers/routers.go
@@ -293,6 +293,24 @@ func NewAPIRouter(han *controllers.APIController, authMiddleware, initMiddleware
 	apiRouter.Handle("/repositories/{repoID}/instances", http.HandlerFunc(han.ListRepoInstancesHandler)).Methods("GET", "OPTIONS")
 	apiRouter.Handle("/repositories/{owner}/{repo}/instances", http.HandlerFunc(han.ListRepoInstancesByFriendlyNameHandler)).Methods("GET", "OPTIONS")
 
+	if manageWebhooks {
+		// Install Webhook
+		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.InstallRepoWebhookHandler)).Methods("POST", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.InstallRepoWebhookByFriendlyNameHandler)).Methods("POST", "OPTIONS")
+		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.InstallRepoWebhookHandler)).Methods("POST", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.InstallRepoWebhookByFriendlyNameHandler)).Methods("POST", "OPTIONS")
+		// Uninstall Webhook
+		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.UninstallRepoWebhookHandler)).Methods("DELETE", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.UninstallRepoWebhookByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
+		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.UninstallRepoWebhookHandler)).Methods("DELETE", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.UninstallRepoWebhookByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
+		// Get webhook info
+		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.GetRepoWebhookInfoHandler)).Methods("GET", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.GetRepoWebhookInfoByFriendlyNameHandler)).Methods("GET", "OPTIONS")
+		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.GetRepoWebhookInfoHandler)).Methods("GET", "OPTIONS")
+		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.GetRepoWebhookInfoByFriendlyNameHandler)).Methods("GET", "OPTIONS")
+	}
+
 	// Get repo
 	apiRouter.Handle("/repositories/{repoID}/", http.HandlerFunc(han.GetRepoByFriendlyNameHandler)).Methods("GET", "OPTIONS")
 	apiRouter.Handle("/repositories/{owner}/{repo}/", http.HandlerFunc(han.GetRepoByIDHandler)).Methods("GET", "OPTIONS")
@@ -315,23 +333,6 @@ func NewAPIRouter(han *controllers.APIController, authMiddleware, initMiddleware
 	apiRouter.Handle("/repositories/", http.HandlerFunc(han.CreateRepoHandler)).Methods("POST", "OPTIONS")
 	apiRouter.Handle("/repositories", http.HandlerFunc(han.CreateRepoHandler)).Methods("POST", "OPTIONS")
 
-	if manageWebhooks {
-		// Install Webhook
-		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.InstallRepoWebhookHandler)).Methods("POST", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.InstallRepoWebhookByFriendlyNameHandler)).Methods("POST", "OPTIONS")
-		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.InstallRepoWebhookHandler)).Methods("POST", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.InstallRepoWebhookByFriendlyNameHandler)).Methods("POST", "OPTIONS")
-		// Uninstall Webhook
-		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.UninstallRepoWebhookHandler)).Methods("DELETE", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.UninstallRepoWebhookByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
-		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.UninstallRepoWebhookHandler)).Methods("DELETE", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.UninstallRepoWebhookByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
-		// Get webhook info
-		apiRouter.Handle("/repositories/{repoID}/webhook/", http.HandlerFunc(han.GetRepoWebhookInfoHandler)).Methods("GET", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook/", http.HandlerFunc(han.GetRepoWebhookInfoByFriendlyNameHandler)).Methods("GET", "OPTIONS")
-		apiRouter.Handle("/repositories/{repoID}/webhook", http.HandlerFunc(han.GetRepoWebhookInfoHandler)).Methods("GET", "OPTIONS")
-		apiRouter.Handle("/repositories/{owner}/{repo}/webhook", http.HandlerFunc(han.GetRepoWebhookInfoByFriendlyNameHandler)).Methods("GET", "OPTIONS")
-	}
 	/////////////////////////////
 	// Organizations and pools //
 	/////////////////////////////


gabriel-samfira:
Rather than escape and unescape owner/repo, the client can split by `/` and make a request to get `/api/v1/repositories/owner/repo-name`. This can be a new URL in the router with a separate handler. For orgs and enterprises you only have the `name`. The new URL can have a query arg `?forgeType=` in case of ambiguity.

This would limit to the client the need to parse things, and the API can have a clear route that deals with entities by name, while still allowing for a `forgeType` to be specified.

What do you think?

gabriel-samfira:
for pools you can try to parse the argument as a UUID. If it fails, we fall back to friendly name parsing. Decide which API call to make based on the argument, in the client.

gabriel-samfira:
To regenerate the swagger code, you can run:

```bash
go generate ./...
```

If you add a new type, you should also update: https://github.com/cloudbase/garm/blob/main/apiserver/swagger.yaml

apiserver/controllers/instances.go
```diff
@@ -228,20 +228,13 @@ func (a *APIController) DeleteInstanceHandler(w http.ResponseWriter, r *http.Req
 //	  default: APIErrorResponse
 func (a *APIController) ListRepoInstancesHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
-	vars := mux.Vars(r)
-	repoID, ok := vars["repoID"]
+
+	repo, ok := a.GetRepository(w, r)
```
gabriel-samfira, line 232: It's okay to have some duplicate code, if we can skip doing an extra query. There are people using GARM at scale (hundreds of runners per second in some cases), adding extra queries increases DB queries and load.

It's also desirable to have separate functions for getting by ID versus getting by owner/name. Branching removes duplication at the expense of clarity and flexibility.

So having a `ListInstancesForRepositoryByNameHandler()`, which has a specific route that is tied to it, is okay. It also allows us to refactor it later without touching the bits that use the ID instead of the owner/name.

It also means having a distinct type in the client code generated by swagger. We want to keep the ability to query by ID alone from the client.

So in summary:

* Existing routes and handlers can remain unchanged
* New handlers and routes can be added to handle the cases where names are used instead of IDs
* For the handlers with names, we also need to be able to specify a query arg indicating `forgeType` and/or `endpointName` in case of multiple results.

gabriel-samfira: You make good points. My apologies for the long reply. I like in depth discussions on some things, so bare with me (I wouldn't blame you if you just `/dev/null` the whole thing :sweat_smile:).

> The swagger duplication was an adventure.

It essentially requires a new handler for each route that wants a friendly name variant. The swagger comments are mostly identical, but for the route and the name of the route (used to generate client code). Parameters also have to be declared. But otherwise, it should be straight forward. It's actually more work to de-duplicate than it is to add new handlers.

> e.g. The shared code had a bug, now I could just fix it everywhere by updating a single line

The reverse is also true. You can add a bug everywhere by changing a single line. And it's much easier to do, when it becomes difficult to determine the consequences of a change throughout the different branches of code (you need to be intimately familiar with the code to understand how one change, cascades through deeply branched code - I've had this happen on more complex code bases).

> I have no idea how you can maintain the duplicated code

It's not that hard. Handlers are atomic. I need to modify one, when I need to change the way a certain endpoint functions. If I don't, there is no point in touching that code. There are places where de-duplicating is great. There are places where duplication is okay. In presentation code like handlers, some duplication is okay. If you need to do a switch in your controller based on which parameters are present, and you delegate to specific functions the processing of that request, the only thing you're de-duplicating is the swagger stuff. And even that becomes ambiguous when transposed into a generated client library.

It's sometimes difficult to decide when it's okay to duplicate and when it's not, but de-duplicating code is much easier to do than splitting up concerns later on, when the code base becomes spaghetti and you can't make heads or tails of what could break if you make a change somewhere.

There is also the matter of allowing for things to diverge in terms of functionality. Let's take the example of getting a repository. In our scenario, if we use the UUID, there is no ambiguity. The ID is unique and it identifies the database entity without issues. If we want to get by name however, we need to take into account the following:

* A repository may have the same owner/name on several endpoints
* Endpoints can be of different types
* There is the potential that in the future, GARM may have the ability to create users which will have their own endpoints defined
  * Admin users should be able to view other user defined repos

So when filtering by name, we will probably need to allow API calls to send additional query filters like endpoint, endpoint type, user (if you're an admin, you should specify the user that created the entity), etc. But if you use the same handler for the case where an ID is used, your client still allows setting additional filters for that case. Should they cause an error? Should they be ignored? How would the docs for the client look like?

An API endpoint that filters only by ID will most likely never change. It will try to match that ID against the database and will either return the entity or a 404 or 401 error.

An API endpoint that filters by name, needs to take into account multiple factors, which may change over time. This is due to the inherent ambiguity of filtering by something that is not unique. We need to supply a set of parameters that together, guarantee uniqueness. That set of parameters may change over time (unlike the UUID) based on features we may need to add (perhaps multi user support?).

So from an architectural stand point, we have to weigh the benefits of not having an eye bleed, now, due to looking at duplicate code and constantly fighting the urge to hack it with an axe, against future growth where we can keep what is currently functional, and only touch what we care about, without having to think about breaking anything else.

So you have a `GetRepoByIDHandler()` that services `/api/v1/repositories/{repoID}` (with no other options) and a `GetRepoByFriendlyNameHandler()` that services `/api/v1/repositories/{owner}/{name}` with potential query args that may change over time. The current one may never change. The other one may be changed independently.

I'd much rather have proper test cases for new code (-*sobs in little time to add tests*-) if the fear is that by duplicating code, we end up with different behavior between old and new code.

All of that being said, I know time is precious (and I am humbled any time someone chooses to use theirs to help). I do not want to burden you with implementing a feature that may have seemed simple to begin with but due to my preferences, or quirks or whatnot, you end up spending a lot of time adding. If you prefer, I can pick this up sometimes in the next 2-3 weeks (if it's not a blocker for you). The user story you presented here is clear (and much easier to use from a UX perspective).

gabriel-samfira: > Doesn't this PR has new client side handlers and request parameters?

If the extra swagger comment generates the proper client, it's fine. 

The initial comment on this line, was due to the fact that there was a call to `a.GetRepository()` which queries the DB for the repo, and then gets the ID of the repo to pass on to `ListRepoInstances()`, where before we got the ID from the route. That has changed now from what I see :smile:, so I will have a closer look when the branch is marked as ready for review.

gabriel-samfira:
> There are multiple options for you to choose from, since I didn't force push.

my brain is mush (had too little sleep), but I will have a look soon.

> How can I run intergration tests locally on ubuntu 25.04 (older ubuntu arm64 didn't had desktop isos for my VM)?

That's a bit more complicated, because the integration tests need access to a test org and a test repo. Moreover, due to the nature of runners, I needed a way to expose the GARM API to github so we can receive webhooks. All this makes integration tests more complicated, at least when coupled to github. For Gitea, it should be way easier to implement tbh.

For creating the reverse proxy/tunnel I created a quick and dirty service (probably full of bugs given it was a 1 day project) here:

https://github.com/gabriel-samfira/localshow

To access this, the current integration job needs a secret that contains a SSH private key. Details bellow. But you can also use ngrok (I think there's an action to set it up) or serveo, or [FRP](https://github.com/fatedier/frp), or [rathole](https://github.com/rapiz1/rathole) (not sure if it's maintained anymore). The idea is to temporarily expose the API from behind a NAT somewhere (typical runner setup).

The needed secrets are:

* GH_OAUTH_TOKEN - this is a PAT with access to the test org and test repo (somewhere on GitHub)
* LXD_REMOTE_SERVER - this is a remote that you can set for faster LXD image downloads. But it's optional. It defaults to https://cloud-images.ubuntu.com/releases
* SSH_PRIVATE_KEY -  the private key used by the tests to create the reverse proxy tunnel (it uses SSH as a transport).
* TUNNEL_BASE_URL - the base URL of the tunnel hosting service. Usually it's something like `tunnel.example.com`, to which the tests generate a random subdomain and you end up with a tunnel like `someRandomSubdomain.tunnel.example.com`.

The last 2 depend on whether you use the quick and dirty reverse proxy I linked to or not. If you go with something that is actually maintained, you can skip those and just update the job. For gitea you don't even need this. Everything can be on the same machine. Gitea in a container, GARM started up in the background by the tests.

I attempted to make it easier to run locally, but the fact that we need github to be able to send us webhooks to make the tests relevant, makes this tough.

gabriel-samfira:
> There are multiple options for you to choose from, since I didn't force push.

TL;DR; the second option would be desirable

My general preference is as follows:

* For presentation to the user, I would rather duplicate some code **if** that makes the interface clearer. 
* Deduplication is preferred whenever possible
* Deduplication is not preferred if it makes the code hard to read/follow/change later.

In the above case, it's okay to duplicate the handlers and have one Swagger definition per handler, but de-duplicate any common code that can be used by both the get by ID and by friendly name.

apiserver/controllers/common.go
```diff
@@ -0,0 +1,57 @@
...
+			Error:   "Bad Request",
+			Details: "No repo ID specified",
+		}); err != nil {
+			slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "failed to encode response")
+		}
+		return params.Repository{}, false
+	}
+	repoObj, err := a.r.ResolveRepository(r.Context(), owner, repo, r.URL.Query().Get("endpointName"))
+	if err != nil {
+		slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "listing pools")
```
gabriel-samfira, line 29:
```suggestion
		slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "resolving repository")
```

apiserver/controllers/common.go
```diff
@@ -0,0 +1,57 @@
...
+			Error:   "Bad Request",
+			Details: "No repo ID specified",
+		}); err != nil {
+			slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "failed to encode response")
+		}
+		return "", false
+	}
+	repoObj, err := a.r.ResolveRepository(r.Context(), owner, repo, r.URL.Query().Get("endpointName"))
+	if err != nil {
+		slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "listing pools")
```
gabriel-samfira, line 52:
```suggestion
		slog.With(slog.Any("error", err)).ErrorContext(r.Context(), "resolving repository")
```

test/integration/client_utils.go
```diff
@@ -214,6 +214,16 @@ func updateRepo(apiCli *client.GarmAPI, apiAuthToken runtime.ClientAuthInfoWrite
 	return &updateRepoResponse.Payload, nil
 }
 
+func updateRepoByName(apiCli *client.GarmAPI, apiAuthToken runtime.ClientAuthInfoWriter, owner, repo string, repoParams params.UpdateEntityParams) (*params.Repository, error) {
+	updateRepoResponse, err := apiCli.Repositories.UpdateRepoByName(
```
gabriel-samfira, line 218: You need to add a route in `routed.go` before this will work.

apiserver/routers/routers.go
```diff
@@ -251,19 +251,29 @@ func NewAPIRouter(han *controllers.APIController, authMiddleware, initMiddleware
 	/////////////////////
 	// Get pool
 	apiRouter.Handle("/repositories/{repoID}/pools/{poolID}/", http.HandlerFunc(han.GetRepoPoolHandler)).Methods("GET", "OPTIONS")
+	apiRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}/", http.HandlerFunc(han.GetRepoByNamePoolHandler)).Methods("GET", "OPTIONS")
```
gabriel-samfira, line 254: You probably need to add routes for CRUD for repositories as well, not just for repo pools. Same for repo instances.

gabriel-samfira (changes requested):
Looks good for the most part. Just a few comments and a few things missing.

Here is an integration test run: https://github.com/gabriel-samfira/garm/actions/runs/15487918541/job/43606694806

Sorry it's taking so long to review.

apiserver/routers/routers.go
gabriel-samfira: That's ok. It's what reviews and tests are for.

apiserver/controllers/repositories.go
```diff
@@ -722,3 +722,710 @@ func (a *APIController) GetRepoWebhookInfoHandler(w http.ResponseWriter, r *http
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(pools); err != nil {
+		slog.With(slog.Any("error", err)).ErrorContext(ctx, "failed to encode response")
+	}
+}
+
+// swagger:route GET /repositories/{owner}/{repo}/scalesets Repositories scalesets ListRepoByNameScaleSets
```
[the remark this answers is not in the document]
gabriel-samfira: > _Naming things is not my strength_

We share that quality.

> If we name this like ListRepoScaleSetsByFriendlyName, then I would expect that I am able to use a friendly name of a scaleset as well instead of a scaleset id.

that's true, and is something we will probably do in the future (I would like to name pools as well at some point), but given that pools (right now at least) can only be referenced by ID, I think it would be sufficiently intuitive that consistency is maintained between pools and scale sets, and should not cause confusion.

When we do add the ability to reference scale sets by name, we can keep the name, handler and URL the same and add aditional query arg options to be able to specify `runnerGroup`  (scale set names are unique within a runner group) and `endpoint`.

gabriel-samfira:
> Is there another big flaw in the routing here?

There is a potential of conflict if there is ever a `webhook` repo under an `{owner}`.

> Growing apis like this is error prone, do we really want to go this way?

We have a few options here, but I need to have the time and mental clarity to think about this and forward a discussion. Growing the API will eventually lead to conflicts, but we have a couple of options we need to weigh the pros and cons of:

1) Doing this client side.
  * Pros: no changes needed to the backend.
  * Cons: adds multiple API calls (minimum 2, potentially more)
2) Using query args instead of URL parameters to feed in the owner, name and optional endpoint
  * Pros: It might work (is this even a pro? - going back under my rock)
  * Cons: Not very REST-e for anything other than `READ`, feels like trying to fit a square in a circle.
3) adding an `/api/v1.1` base route for new functionality.
  * Pros: Allows for growth, clean implementation, old clients continue to function unchanged, new clients know where to request the new functionality
  * Cons: API versioning requires discipline

There might be other options, but I can't allocate time now to proper thought/research.

So, to conclude. You're absolutely right that this can't grow much further without asking for trouble. The conflict you identified is pretty glaring. If you think that at this point it's easier to have a CLI-only feature that emulates what you tried to add with this PR, I am all for it (with a tracking issue opened to potentially implement this server side in the future, via an option that is clean and correct - which can be implemented by whomever has the cycles).

What are your thoughts on this? I am open to anything.

gabriel-samfira:
I think it's possible. Worth a try. It might be cleaner to version the API, though. Something like:

```go
apiV11SubRouter := router.PathPrefix("/api/v1.1").Subrouter()
apiV11SubRouter.Use(initMiddleware.Middleware)
apiV11SubRouter.Use(urlsRequiredMiddleware.Middleware)
apiV11SubRouter.Use(authMiddleware.Middleware)
apiV11SubRouter.Use(auth.AdminRequiredMiddleware)

apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}/", http.HandlerFunc(han.GetRepoPoolByFriendlyNameHandler)).Methods("GET", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}", http.HandlerFunc(han.GetRepoPoolByFriendlyNameHandler)).Methods("GET", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}/", http.HandlerFunc(han.DeleteRepoPoolByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}", http.HandlerFunc(han.DeleteRepoPoolByFriendlyNameHandler)).Methods("DELETE", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}/", http.HandlerFunc(han.UpdateRepoPoolByFriendlyNameHandler)).Methods("PUT", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/{poolID}", http.HandlerFunc(han.UpdateRepoPoolByFriendlyNameHandler)).Methods("PUT", "OPTIONS")
apiV11SubRouter.Handle("/repositories/{owner}/{repo}/pools/", http.HandlerFunc(han.ListRepoPoolsByFriendlyNameHandler)).Methods("GET", "OPTIONS")

................ The rest of the friendly name URLs ................
```

The only new code in the example above is defining the new `/api/v1.1` prefix (first 5 lines). The rest is copy paste from your PR, with `apiRouter` substituted for `apiV11SubRouter`.

Regex might also work for the UUID. I mean, if someone names their repository using `uuidgen`..... :shrug:. There are some things we just don't want to account for tbh. So either way I think is fine.

gabriel-samfira:
> Or doing something similar to `/repositories/byFriendlyName/<owner>/<repo>/...` and `/repositories/byID/<owner>/<repo>/...` in v1.1 only?

The old router would still exist and we don't need to duplicate old routes that exist there, in `v1.1`. I am thinking of a union between old and new versions. Hypothetically, say your client wants to run an operation that lists repo pools by friendly name. The client can check the "minimum API version" for that operation and make sure to send that operation against the proper API prefix. 

If the client needs to list pools by ID, it knows the minimum API version for that is `v1`. So the endpoint can be found there.

So only the friendly name URLs would need to exist (currently) in that router. Whatever exists now in `v1` would stay there and continue to function as it does now. But this is just an idea.

> I assume we agree calling RepoList to query the id is not good. So I assume we need at least one new backend endpoint.

Yes. I suggest adding a `filter` option to the list repo/org/enterprise endpoints that allows you to do something like:

```
GET /api/v1/repositories/?owner=repoOwner&name=repoName&endpoint=repoEndpoint
```

The list repo/org/enterprise handler can look for the query args and adjust the DB query accordingly. This should allow us to return a small subset (min 0, max depending on whether or not `endpoint=` was set and the number of endpoints that exist with the same combination of owner/repo name).

So either of these 2 approaches are ok. The client side one with minimal changes to the list entity endpoints seems easier. The API versioning one needs more thought and careful planning.

PR #431, opened 2025-06-17
Add a rudimentary filter option when listing entities

gabriel-samfira:
This change adds the ability to filter the list of entities returned by the API by entity owner, name or endpoint, depending on the entity type.

Example:

[asciicast](https://asciinema.org/a/gXUSqQSkZiBpujGfIbCHlZhwN)


gabriel-samfira:
CC @ChristopherHX 

Re #421

PR #444, opened 2025-07-02
improve usability of podman in Makefile

chickenkiller:
The previous syntax worked with a real shim file 'docker' pointing to podman (or a symlink). This P/R makes use of podman even when 'docker' symlink or alias is absent.


gabriel-samfira:
thanks!

PR #445, opened 2025-07-06
Auto increment patch dirty version

ChristopherHX:
* resolve semver comparing problems in garm-operator

* if you checkout the exact tag and do make build => version = `<tagname>`
* if you edit a file in the folder => `<tagname, patch incremented by 1>-dirty`
* if you commit the edit to the file in the folder => `<tagname, patch incremented by 1>-<no>-<shortsha>`
Before `<tagname>` was

> I need to cut v0.1.6, but that will happen from the release/v0.1 branch, which will now diverge from main, which will become release/v0.2 as soon as I have time. The nightly build will then report the version as being v0.2.0-000-g0000.

Not until you do tag v2.0.0 and the problem is still there? Or do I miss something.

I mean if make edits and you just run `make build`, I would expect a tag greater than the tag checked out.

A lot of programs consider `<tagname>-dirty` < `<tagname>` and will fail if they require `<tagname>`. Given someone making edits would not expect weird problems due to the version ordering.

See <https://github.com/mercedes-benz/garm-operator/issues/285>

CC @sailorbob134280

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ USER_ID=$(shell ((docker --version | grep -q podman) && echo "0" || id -u))
 USER_GROUP=$(shell ((docker --version | grep -q podman) && echo "0" || id -g))
 ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 GOPATH ?= $(shell go env GOPATH)
-VERSION ?= $(shell git describe --tags --match='v[0-9]*' --dirty --always)
+VERSION ?= $(shell ./scripts/get-version.sh)
 GARM_REF ?= $(shell git rev-parse --abbrev-ref HEAD)
 GO ?= go
 export GARM_PASSWORD ?= ${GEN_PASSWORD}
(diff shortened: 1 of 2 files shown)


gabriel-samfira:
I think this can be fixed with better house keeping. And that is my responsibility. After any new release, I need to immediately create a new tag that increments the minor version and adds a `-alpha1` or something. This is something I need to do, not something that should be fixed using code.

Let me take care of that this evening. I need to tag `v0.1.6` from the `release/v0.1` branch anyway, I will also create the `release/v0.2` branch and also tag `v0.2.0-alpha1` from `main`. And this issue will go away.

What do you think?

gabriel-samfira:
that is the plan. The release process would look something like:

* Tag a new release from HEAD
* Push the tag
* Update README.md to reflect that main now tracks the new alpha version
* Tag the new alpha version that increments the minor version and adds an `-alpha1`
* Push new tag

At which point, anyone that clones `main` should get the new alpha tag when building from it.

gabriel-samfira:
> Closed as 50% fixed

Please feel free to open an issue for the other 50%. I will try to address whatever is missing. Also, if you get stuck with various things, you can always ping me on slack. I will try to help and hopefully save you some time.

PR #456, opened 2025-07-17
Refactor CI/CD builds

chickenkiller:
This P/R makes the workflows a bit cleaner and enable the build and push of 2 versions of garm images.

* The providers branch computation is done inside the Dockerfile now
* It automatically computes the latest (non pre-release) release of each provider before cloning

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,14 +1,16 @@
-name: "Build GARM images"
+name: "Build and push GARM images"
 on:
-  workflow_dispatch:
+  workflow_call:
     inputs:
       push_to_project:
         description: "Project to build images for"
         required: true
+        type: string
         default: "ghcr.io/cloudbase"
       ref:
         description: "Ref to build"
         required: true
+        type: string
         default: "main"
   schedule:
     - cron: "0 2 * * *"
@@ -24,8 +26,9 @@ jobs:
       runs-on: ubuntu-latest
       steps:
         - name: "Checkout"
-          uses: actions/checkout@v3
+          uses: actions/checkout@v4
           with:
+            ref: ${{ inputs.ref }}
             path: src/github.com/cloudbase/garm
             fetch-depth: 0
 
@@ -40,15 +43,12 @@ jobs:
             password: ${{ secrets.GITHUB_TOKEN }}
 
         - name: Build and push
+          env:
+            REGISTRY_INPUT: ${{ inputs.push_to_project }}
+            GH_REF: ${{ inputs.ref }}
+          working-directory: src/github.com/cloudbase/garm
           run: |
             set -x
-            REGISTRY_INPUT="${{ github.event.inputs.push_to_project }}"
-            REF_INPUT="${{ github.event.inputs.ref }}"
-
-            PUSH_TO_PROJECT="${REGISTRY_INPUT:-ghcr.io/cloudbase}"
-            GH_REF="${REF_INPUT:-main}"
-            cd src/github.com/cloudbase/garm && git checkout "${GH_REF}"
-
             VERSION=$(git describe --tags --match='v[0-9]*' --always)
             AZURE_REF=v0.1.0
             OPENSTACK_REF=v0.1.0
@@ -68,6 +68,9 @@ jobs:
               EQUINIX_REF="main"
               K8S_REF="main"
               VERSION="nightly"
+            fi
+            if [ "$GH_REF" == "release/v1" ]; then
+              VERSION="v0.1"
             fi
             docker buildx build \
               --provenance=false \
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -48,7 +48,6 @@ jobs:
             GH_REF: ${{ inputs.ref }}
           working-directory: src/github.com/cloudbase/garm
           run: |
-            set -x
             if [ "$GH_REF" == "main" ]; then
               VERSION="nightly"
             else
@@ -67,5 +66,5 @@ jobs:
               --label "org.opencontainers.image.description=GARM ${GH_REF}" \
               --label "org.opencontainers.image.licenses=Apache 2.0" \
               --build-arg="GARM_REF=${GH_REF}" \
-              -t ${PUSH_TO_PROJECT}/garm:"${VERSION}" \
+              -t ${REGISTRY_INPUT}/garm:"${VERSION}" \
               --push .
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -78,10 +78,10 @@ jobs:
               LINODE_REF=$(get_gh_latest_release flatcar/garm-provider-linode)
               K8S_REF=$(get_gh_latest_release mercedes-benz/garm-provider-k8s)
             fi
-            if [ "$GH_REF" == "release/v1" ]; then
+            if [ "$GH_REF" == "release/v0.1" ]; then
               VERSION="v0.1"
             fi
-            if [ "$GH_REF" == "release/v2" ]; then
+            if [ "$GH_REF" == "release/v0.2" ]; then
               VERSION="v0.2"
             fi
             docker buildx build \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -32,9 +33,9 @@ default: build
 .PHONY : build-static test install-lint-deps lint go-test fmt fmtcheck verify-vendor verify create-release-files release
 build-static: ## Build garm statically
 	@echo Building garm
-	docker build --tag $(IMAGE_TAG) -f Dockerfile.build-static .
+	$(IMAGE_BUILDER) build -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt --tag $(IMAGE_TAG) -f Dockerfile.build-static .
```
gabriel-samfira, line 36: This will most likely break docker. See:

https://github.com/cloudbase/garm/blob/bb45324dcd929193ec31a06bcbc450a114668fb4/Makefile#L25-L27

What would probably work on both podman and docker (provided podman implements the `--mount` option for thr `RUN` stanza is changing the Dockerfile with something like this:

```Dockerfile
RUN --mount=type=bind,source=/etc/ssl/certs/ca-certificates.crt,target=/etc/ssl/certs/ca-certificates.crt REST_OF_COMMANDS_GO_HERE
```

.github/workflows/build-and-push.yml
```diff
@@ -38,15 +41,49 @@ jobs:
...
+          env:
+            REGISTRY_INPUT: ${{ inputs.push_to_project }}
+            GH_REF: ${{ inputs.ref }}
+          working-directory: src/github.com/cloudbase/garm
           run: |
-            cd src/github.com/cloudbase/garm
-            VERSION=$(git describe --tags --match='v[0-9]*' --always ${{ github.event.inputs.ref }})
+            set -x
+            VERSION=$(git describe --tags --match='v[0-9]*' --always)
+            AZURE_REF=v0.1.0
```
gabriel-samfira, line 51: I've been thinking about how we could avoid having to hardcode the version numbers for stable releases. One of the reasons to build an image more frequently is to also update the providers. I think we can get the current releases of any github repo, programatically using something like this:

```bash
curl -L   \
    -H "Accept: application/vnd.github+json"   \
    -H "X-GitHub-Api-Version: 2022-11-28"  \
    https://api.github.com/repos/cloudbase/garm/releases | jq '.[] | {name: .name, tag_name: .tag_name, prerelease: .prerelease, created_at: .created_at}'
```

for building from `main` we would still hardcode main.

gabriel-samfira: We'd need to look for the `prerelease` field and still compare semantic versions to make sure that we are indeed building the latest stable and not a pre-release.

Makefile
chickenkiller: Damn, this change was not expected to be included in this P/R. I will fix that.

.github/workflows/build-and-push.yml
gabriel-samfira: for stable releases we need to use actual releases of providers. At some point, I hope we can find a better wy to distribute providers rather than pack them into the garm container.

For "nightly" builds, it's okay to use the `main` branch to build the providers. Nightlies are supposed to be unstable to help testing.

But this is all off the top of my head. What do you think? I am open to any suggestion.

chickenkiller: Your suggestion is good, using latest releases of providers for stable garm releases, and use main branch for nightly.

.github/workflows/build-and-push.yml
```diff
@@ -38,15 +41,61 @@ jobs:
...
+              OPENSTACK_REF=$(get_gh_latest_release cloudbase/garm-provider-openstack)
+              LXD_REF=$(get_gh_latest_release cloudbase/garm-provider-lxd)
+              INCUS_REF=$(get_gh_latest_release cloudbase/garm-provider-incus)
+              AWS_REF=$(get_gh_latest_release cloudbase/garm-provider-aws)
+              GCP_REF=$(get_gh_latest_release cloudbase/garm-provider-gcp)
+              EQUINIX_REF=$(get_gh_latest_release cloudbase/garm-provider-equinix)
+              LINODE_REF=$(get_gh_latest_release flatcar/garm-provider-linode)
+              K8S_REF=$(get_gh_latest_release mercedes-benz/garm-provider-k8s)
+            fi
+            if [ "$GH_REF" == "release/v1" ]; then
```
gabriel-samfira, line 79: The release branches are called: `release/v0.1` and `release/v0.2`. You can see the names of the branches we merge into before release, in the "switch branches/tags" dropdown menu.

gabriel-samfira (changes requested):
This is awesome! Looks much cleaner. I have a few comments though. And also, do you think it makes sense to propose this against the `main` branch? I am not sure it needs to be merged only in the `release/v0.1` branch, considering it can build for `main`, `release/v0.1` and `release/v0.2`.

Dockerfile
```diff
@@ -1,48 +1,56 @@
...
+      cloudbase/garm-provider-aws \
+      cloudbase/garm-provider-gcp \
+      cloudbase/garm-provider-equinix \
+      flatcar/garm-provider-linode \
+      mercedes-benz/garm-provider-k8s; \
+    do \
+        export PROVIDER_NAME="$(basename $repo)"; \
+        export PROVIDER_SUBDIR=""; \
+        if [ "$GARM_REF" == "main" ]; then \
+          export PROVIDER_TAG="main"; \
```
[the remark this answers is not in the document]
chickenkiller: yes I had a doubt about passing `main` as the `main.Version` value but I wanted to be functionally equivalent to before. I will work on that.

Dockerfile
```diff
@@ -1,48 +1,56 @@
...
+        else \
+          export PROVIDER_TAG="$(curl -s -L https://api.github.com/repos/$repo/releases/latest | jq -r '.tag_name')"; \
+        fi; \
+        case $PROVIDER_NAME in \
+        "garm-provider-k8s") \
+            export PROVIDER_TAG=v0.3.1; \
```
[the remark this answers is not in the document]
chickenkiller: It was hardcoded in Dockerfile, I thought there was a reason so I added the override here too. Let's talk with the provider team if needed and change this if we can.

.github/workflows/build-and-push.yml
```diff
@@ -38,15 +41,28 @@ jobs:
...
           run: |
-            cd src/github.com/cloudbase/garm
-            VERSION=$(git describe --tags --match='v[0-9]*' --always ${{ github.event.inputs.ref }})
+            if [ "$GH_REF" == "main" ]; then
+              VERSION="nightly"
+            else
+              VERSION=$(git describe --tags --match='v[0-9]*' --always)
+            fi
+            if [ "$GH_REF" == "release/v0.1" ]; then
+              VERSION="v0.1"
```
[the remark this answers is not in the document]
gabriel-samfira: both `release/v0.1` and `release/v0.2` will release "stable" versions. It's just that `v0.2.x` is in theory a major release (I need to cut v1 at some point to make things easier) and has features/changes that are breaking from `v0.1`. For example, the garm operator works with `v0.1.6`. I would like to no longer break it (or any other integrations) for minor releases. So I would like for the `release/v0.1` branch to only receive bug fixes and nothing that changes the API.

Eventually, `v0.1` will be retired and no new versions will be released from that. But we need to go in parallel for a while to give people a chance to update and test.

> Another solution is to call the tag v0.1.x instead of v0.1

As a general housekeeping rule, whenever I release a new stable version, the `release/v0.x` branch is incremented by 1 and an `-alpha` suffix is added, until we want to release a new version and we remove the suffix. So if we just released `v0.1.6`, we would also add a new commit to the `release/v0.1` branch and tag that as `v0.1.7-alpha`. This way we can merge new things in the release branch and the version will appear to be newer.

I aim to do this for all branches (main included). So if we use `git describe --tags`, we should get a unique, incremental version. I am not sure though how to best mark the container image as "latest"/"nightly" relative to the branch it was cur from (latest v0.1, latest v0.2, etc). Not sure if it's important, either. In general, I would advise people to use stable releases. The "nightly" is for testing and can explode.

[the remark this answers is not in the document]
gabriel-samfira: I am trying to figure out how we could best do this, and keep it as least confusing as possible.

PR #489, opened 2025-08-19
Feat: Add Helm Chart

igrikus:
First off, thank you for this amazing project! It's been incredibly helpful for my work

I prefer to manage all my applications with Kubernetes, so I've created a Helm chart that can help with that

**Key Features:**
- Performs `garm-cli init` using either an auto-generated password or a pre-existing secret.

- Registers one or more Gitea/GitHub endpoints and credentials using a pre-existing secret.

- Dynamically registers one or more cloud providers.

- Post-install configuration (like adding repositories and runner pools) is streamlined for use with your new UI, accessible via an Ingress resource.

**Points to Consider:**
- To make this chart easily consumable for end-users, it should be published to a repository like _Artifact Hub_. If you decide to merge this, I'd be happy to help with the next steps to get it ready for publication.

- I was unable to test this on a GitHub Enterprise instance as I don't have access to one.

- My testing was limited to a Gitea/GitHub + GCP setup, so there might be unhandled edge cases for other cloud providers that require additional configuration

- I'm not certain about the best placement for the chart files within the project structure. I am completely open to your suggestions, whether it's moving them to a dedicated `charts/` directory or into a **separate repository** altogether.

### DIFF
--- a/helm-chart/templates/configmap-init-script.yaml
+++ b/helm-chart/templates/configmap-init-script.yaml
@@ -26,9 +26,9 @@ data:
       garm-cli init \
         --name "{{ include "garm.fullname" . }}" \
         --url "http://localhost:{{ .Values.service.port }}" \
-        --callback-url "{{- if .Values.garm.callbackUrl }}{{ .Values.garm.callbackUrl }}{{- else }}{{ .Values.garm.url | trimSuffix "/" }}/api/v1/callbacks{{- end }}" \
-        --metadata-url "{{- if .Values.garm.metadataUrl }}{{ .Values.garm.metadataUrl }}{{- else }}{{ .Values.garm.url | trimSuffix "/" }}/api/v1/metadata{{- end }}" \
-        --webhook-url "{{- if .Values.garm.webhookUrl }}{{ .Values.garm.webhookUrl }}{{- else }}{{ .Values.garm.url | trimSuffix "/" }}/webhooks{{- end }}" \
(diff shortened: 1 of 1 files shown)


gabriel-samfira:
Hi @igrikus 

This is great!

However, I think at this point, the best place for a helm chart is in the k8s operator. While I love the idea of having a stand-alone GARM instance in k8s, the reality is that most people that will deploy it on k8s, will want to use the same experience they use when interacting with k8s to configure GARM itself. And this is where the operator comes in. I think there's already an issue open on the matter in the operator repo.

Another reason I never added a helm chart for GARM itself is that it changes a lot from version to version and it's more difficult to maintain than if it were a part of a project like the k8s operator (which tracks stable versions).

But for those that do want to set up GARM as a stand alone service in k8s, I think that this can be maintained as a separate project (if you're willing).

I would also love to encourage a community to grow around GARM rather than concentrate everything in this repo. The more people use it and help one another to set it up, the better. And the hard work that the community does shouldn't depend on the approval of anyone in this project.

So I would love to see separate repos for providers, helm charts, enablement of any kind with discussions posted either in the discussions section or anywhere else.

What do you think?

igrikus:
Hi @gabriel-samfira 

That's a great point, and I completely agree

Keeping the Helm chart in a separate repository makes perfect sense, especially for maintenance

I definitely can create my own repository for the chart. Once it's ready, would you be open to adding a link to it in the `README.md`, similar to how the k8s operator is mentioned?

With that said, feel free to close this pull request. Thanks for the feedback!

PR #491, opened 2025-08-21
fix(jobs): Correctly handle workflow job IDs from webhooks

igrikus:
### Problem

Previously, `garm` was not automatically creating new runners when new jobs appeared in Gitea. This was happening because the system was getting confused about job identification.

Here's what went wrong:

1.  **Saving Job Data:** When `garm` tried to save information about a new job into its database, it was accidentally using its own internal database ID for the job, instead of the unique job ID provided by Gitea. So,
    the database record ended up with the wrong external ID.
2.  **Looking Up Jobs:** Later, when `garm` received updates about these jobs (like a webhook saying a job started), it tried to find the job in its database. But it was looking for the job using the unique ID from  
    Gitea, and trying to match it against its _own internal database ID_. Since the IDs didn't match (because of the first problem), `garm` couldn't find the job.

Because `garm` couldn't correctly identify the jobs, it failed to queue new runners for them.

### Solution

This change fixes how `garm` handles job IDs:

1.  **Correct Saving:** Now, when job information is saved, `garm` correctly stores the unique job ID from GitHub/Gitea in the right place in its database record.
2.  **Correct Lookup:** When `garm` needs to find a job, it now correctly searches using the unique job ID from GitHub/Gitea against the corresponding field in its database.

These fixes ensure that `garm` can properly understand and track jobs from GitHub/Gitea, allowing it to automatically create and manage runners as expected.

### DIFF
--- a/database/sql/jobs.go
+++ b/database/sql/jobs.go
@@ -248,7 +248,7 @@ func (s *sqlDatabase) CreateOrUpdateJob(ctx context.Context, job params.Job) (pa
 	var err error
 
 	searchField := "workflow_job_id = ?"
-	var searchVal any = job.ID
+	var searchVal any = job.WorkflowJobID
 	if job.ScaleSetJobID != "" {
 		searchField = "scale_set_job_id = ?"
 		searchVal = job.ScaleSetJobID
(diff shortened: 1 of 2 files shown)


gabriel-samfira:
I need to add e2e tests for gitea. Will do that soon.

gabriel-samfira:
Thanks!

PR #562, opened 2025-10-17
Drop empty/redundant components from runner service name

bachorp:
I have observed unit names like `actions.runner.myorg..service`. Is far as I understand, for enterprises and organizations, `Owner` will simultaneously be the entity name while `Name` will be remain empty.


gabriel-samfira (approved):
thanks!

PR #611, opened 2026-02-11
Build webapp assets during Docker build instead of checking them into git

### LINKED ISSUE #609: Don't check in the generated UI code
This makes the diffs huge and is a constant source of conflicts and generally is not good practices.

benoit-nexthop:
This closes #609 

- Add webapp build stage to `Dockerfile` and `Dockerfile.build-static` using `node:lts-alpine`
- Move `build-webapp.sh` to `scripts/` directory
- Add `.dockerignore` to exclude `node_modules` and reduce build context
- Update `.gitignore` to exclude compiled `webapp` assets
- Remove compiled `webapp` assets from repository, keeping it up to date was a constant source of conflict and doing this was bad practice generally anyway


Makefile
gabriel-samfira: Then this section can be replaced with something like:


```Makefile
build-static: ## Build garm statically
	@echo Building garm
	mkdir -p build
	$(IMAGE_BUILDER) build $(EXTRA_ARGS) \
		--build-arg GARM_REF=$(GARM_REF) \
		--output type=local,dest=$(PWD)/build \
		--target=export \
		-f Dockerfile.build-static .
	@echo Binaries are available in $(PWD)/build
```

gabriel-samfira (changes requested):
I've been meaning to address this for a while, but have had very little time lately. So thanks!

Just a few comments for the build-static bit. Let me know what you think

.dockerignore
```diff
@@ -0,0 +1,18 @@
+# Node.js
+webapp/node_modules/
+webapp/build/
+
+# Go
+vendor/
```
gabriel-samfira: This should be removed, otherwise `build-static.sh` will fail. As the build commands there use `-mod vendor`

Makefile
gabriel-samfira: We should probably also add `--no-cache` to the build command.

gabriel-samfira:
will merge as-is and apply the diff in a separate branch

PR #701, opened 2026-04-03
fix: windows ExecutionPolicy in wrapper

brandonkal:
### Summary

   Previously, this userdata wrapper could fail when `Set-ExecutionPolicy` was run in an environment where the effective policy was already more permissive.

   ### Change

   Use:

   ```powershell
   Set-ExecutionPolicy RemoteSigned -Force -ErrorAction SilentlyContinue
 ```

 instead of a plain Set-ExecutionPolicy RemoteSigned.

 ### Why

 This avoids the warning/error path that could stop userdata execution before the install script runs.


gabriel-samfira (approved):
Thank you!

PR #755, opened 2026-06-01
feat(db): add PostgreSQL backend

### LINKED ISSUE #753: Support PostgreSQL as a database backend
### Problem

GARM's only documented and supported database backend is SQLite (the FAQ states "SQLite3 only"; there is some latent, undocumented MySQL code, but no externally-backed backend is supported in practice).  SQLite stores state in a single local file, which ties a GARM deployment to one instance and one persistent volume. This makes it hard to:

- run GARM with an external, managed backing service (rather than a file on a PVC), and
-  move toward running multiple GARM replicas / HA, since the state can't be shared.

### Proposal

 Add PostgreSQL as an alternative database backend, selectable via configuration, alongside the existing SQLite support.
 The database layer already abstracts over multiple SQL backends through gorm (SQLite plus latent MySQL wiring), so PostgreSQL fits as an additional backend = "postgresql" option with no architectural change.

### Motivation 

We're building a Juju charm for GARM (https://github.com/canonical/github-runner-operators) using the 12-factor approach (https://github.com/canonical/paas-charm). 12-factor apps expect external backing services, so a PVC + SQLite doesn't fit that model. We also have in-house experience operating PostgreSQL, including PostgreSQL-as-a-service.

cbartz:
Closes #753 

## What

Add PostgreSQL as a third SQL backend alongside SQLite and MySQL, selectable via
`backend = "postgresql"`:

- **config** — PostgreSQL backend type, connection settings, validation, and DSN
  construction (sslmode, port-range checks, special-character quoting).
- **driver wiring** — open the pgx-backed GORM driver, apply connection-pool
  tuning (max open/idle, lifetimes), and reuse the main connection for blob
  storage (PostgreSQL needs no separate objects database like SQLite uses).
- **file store** — store file-object blobs in PostgreSQL **Large Objects**
  (keyed by `lo_oid`), not a `bytea` column. Large Objects support chunked,
  seekable streaming via the `lo` API, so blobs are read/written incrementally
  and never fully held in memory — required because GARM's file-object interface
  is streaming (`CreateFileObject(..., io.Reader)` /
  `OpenFileObjectContent(...) io.ReadCloser`), and a `bytea` value would have to
  be materialized in full. Writes `io.Copy` the reader into the Large Object
  (hashing inline); reads return a `ReadCloser` over `lo.Read`. SQLite continues
  to use the `Content` BLOB column.
- **migrations** — a dedicated `file_object_migrations` tracking table with a
  rename shim so existing SQLite installs preserve their file-object migration
  history, plus a functional `LOWER(tag)` index and the `lo_oid` column for
  `file_object` blobs.
- **uuid guard** — `getUserByID` pre-validates UUIDs so PostgreSQL's `uuid`
  column returns `ErrNotFound` on malformed input instead of a driver error.
- **tests + CI** — parametrize the SQL suite to run against PostgreSQL when
  `GARM_TEST_POSTGRES_DSN` is set, and add a PostgreSQL job to the test workflow.

## Notes for reviewers

- The bulk of the diff is vendored dependencies (`pgx`, `gorm.io/driver/postgres`
  and transitive deps). The hand-written change is ~970 lines; the rest is
  `vendor/`.
- **Case folding differs subtly between backends:** SQLite's `LOWER()` folds
  ASCII only, PostgreSQL's is locale-aware (folds Unicode). For GARM's
  effectively-ASCII identifiers this is a non-issue, but noting it explicitly.

## Testing

- **Unit/store suite (local, Docker PostgreSQL):** ran the SQL store and config
  suites against `postgres:16` (Docker, `fsync=off`) using the same image,
  credentials, and DSN as the CI job — `database/sql` and `config` pass with
  `-race -tags testing -count=1`.
  ok  github.com/cloudbase/garm/database/sql   (PostgreSQL backend)
  ok  github.com/cloudbase/garm/config
- **SQLite (unchanged backend):** same suite passes in SQLite mode, so both
backends are covered.
- **Lint:** `make lint` (golangci-lint v2.10.1) is clean — 0 issues.
- **CI:** adds a "Go Tests (PostgreSQL)" job that runs `make go-test` against a
`postgres:16` service via `GARM_TEST_POSTGRES_DSN`, exercising
migrations/AutoMigrate and the Large Object file store on a real PostgreSQL.
- **Manual end-to-end:** validated a PostgreSQL-backed GARM end-to-end on
`cbartz/garm` — created credentials, repo, and a scale set, and confirmed
runners register with GitHub and execute jobs.

## Future work

The nightly integration suite (`test/integration`, scheduled in
`integration-tests.yml`) still runs only against SQLite — its config hardcodes
`backend = "sqlite3"`. The PostgreSQL path was validated manually end-to-end (see
Testing); automating that in the nightly suite (add a PG service + template the
`[database]` config) would be a natural follow-up.


gabriel-samfira:
I added docker to the CI runners. You should be able to use the `service:` stanza. If you run into any missing packages, just ping me.

gabriel-samfira (approved):

gabriel-samfira:
Thank you!

Issue #796, opened 2026-06-30
Issue with the data type of template data in the OpenAPI spec


yhaliaw:
In the OpenAPI spec, the `data` field in the template is defined as an array of integers:
```
    Template:
        properties:
            created_at:
                format: date-time
                type: string
                x-go-name: CreatedAt
            data:
                items:
                    format: uint8
                    type: integer
                type: array
                x-go-name: Data
```

When I query it, the datatype seems to be a string:
```
curl -s -H "Authorization: Bearer $TOKEN" http://localhost:9997/api/v1/templates/2
{"id":2,"created_at":"2026-06-03T04:45:47.656934536Z","updated_at":"2026-06-03T04:45:47.656934536Z","name":"github_linux","description":"Default Linux runner install template for GitHub","os_type":"linux","forge_type":"github","data":"IyEvYmluL2Jhc2gKCnNldCAtZQpzZXQgLW8gcGlwZWZhaWwKc2V0I...
```

While Go's `json.Unmarshal` handles this fine, some other languages might not handle this conversion.
The OpenAPI generator for Python uses Pydantic, which fails when validating the received response.

This also applies to the other template-related APIs:
```
    CreateTemplateParams:
        properties:
            data:
                items:
                    format: uint8
                    type: integer
                type: array
                x-go-name: Data
...
    UpdateTemplateParams:
        properties:
            data:
                items:
                    format: uint8
                    type: integer
                type: array
                x-go-name: Data
```

Can the Go code be updated so the OpenAPI spec would set the data field as string?

gabriel-samfira:
can you generate a client using: https://github.com/gabriel-samfira/garm/blob/1391f40a37d658c9f9c0daf147c5f8dc33782ba5/webapp/swagger.yaml and see if that helps? There is a PR that adds a new entity type where this was already added. 

you can use:

```bash
openapi-generator-cli generate --skip-validate-spec -i swagger.yaml -g python -o $HOME/generated
```

where `swagger.yaml` is the file linked to above. Added some swagger annotations on `[]byte` fields which yield:

https://github.com/gabriel-samfira/garm/blob/1391f40a37d658c9f9c0daf147c5f8dc33782ba5/webapp/swagger.yaml#L1972-L1975

PR #800, opened 2026-06-30
Add forge instance entity

gabriel-samfira:
This change adds a new forge entity type along side `Repository{}`, `Organization{}` and `Enterprise{}` called `ForgeInstance{}`. This new entity type is available currently only on `Gitea` and possibly on `Forgejo` if we ever get around to adding it.

The new `ForgeInstance{}` entity requires gitea credentials that have `rw` access to the `admin` api. This feels a bit too much to allow GARM to have, but if instance level pools are required, it cannot be helped.

The API is similar to the other entity types. WebUI has been updated to accomodate the new `ForgeInstance` type, as has the CLI.

PR #801, opened 2026-06-30
Fix/type template data

yhaliaw:
This PR is intended to fix issue #796.

The `byte[]` data type for field `data` in the template causes the OpenAPI spec to record it as an array of uint8, when the returned response is a base64-encoded string.

This PR uses the `Base64` type of `github.com/go-openapi/strfmt` package to ensure the OpenAPI spec generated contains `string` as the datatype.

I have run `make generate` on the code.

### DIFF
--- a/params/requests.go
+++ b/params/requests.go
@@ -25,6 +25,7 @@ import (
 	runnerErrors "github.com/cloudbase/garm-provider-common/errors"
 	commonParams "github.com/cloudbase/garm-provider-common/params"
 	commonUtil "github.com/cloudbase/garm-provider-common/util"
+	"github.com/go-openapi/strfmt"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


params/requests.go
```diff
@@ -877,7 +878,7 @@ func (u UpdateGiteaCredentialsParams) Validate() error {
 type CreateTemplateParams struct {
 	Name        string              `json:"name"`
 	Description string              `json:"description"`
-	Data        []byte              `json:"data"`
+	Data        strfmt.Base64       `json:"data"`
```
gabriel-samfira, line 881: You can just add:

```go
// swagger:strfmt byte
Data        []byte              `json:"data"`
```

and

```go
// go:generate go run github.com/go-swagger/go-swagger/cmd/swagger@v0.33.1 generate spec --output=../swagger.yaml --scan-models --work-dir=../../
```

will do the right thing.

gabriel-samfira: I added the annotations in https://github.com/cloudbase/garm/pull/800. Will probably split that out and merge separately.

gabriel-samfira:
Added a PR to hopefully fix this:

https://github.com/cloudbase/garm/pull/802

yhaliaw:
Closing this since it is covered by another PR.

PR #812, opened 2026-07-07
scaleset: tolerate invalid instance status transitions in job message handlers

benoit-nexthop:
`HandleJobsCompleted` and `HandleJobsStarted` only tolerated `ErrNotFound` when updating an instance's status/runner_status in response to a scale set job message. If the instance had already transitioned to a terminal state (e.g. `deleted`, because the reconciler noticed it went missing from the provider/GitHub after a host failure) by the time a delayed "job started"/"job completed" message arrived, `UpdateInstance` would reject the transition with a `BadRequestError`.

That error was treated as fatal, causing `handleSessionMessage` to return before advancing the message cursor or deleting the message. Since the same poisoned message would then be redelivered forever, the entire scale set listener would get stuck retrying it, blocking every subsequent message (including new job assignments) for that scale set.

Observed in our prod cluster: two scale sets stuck in a tight retry loop logging `invalid instance status transition from deleted to pending_delete` hundreds of times, with a large backlog of queued jobs and idle runners that could never receive them.

This change treats `BadRequestError` from `UpdateInstance` the same way `ErrNotFound` is already treated: log and move on, since it means some other codepath already resolved the runner's terminal state and there's nothing left for us to do.

### DIFF
--- a/workers/scaleset/scaleset_helper.go
+++ b/workers/scaleset/scaleset_helper.go
@@ -122,10 +122,25 @@ func (w *Worker) HandleJobsCompleted(jobs []params.ScaleSetJobMessage) (err erro
 		locking.Lock(job.RunnerName, w.consumerID)
 		_, err := w.store.UpdateInstance(w.ctx, job.RunnerName, runnerUpdateParams)
 		if err != nil {
-			if !errors.Is(err, runnerErrors.ErrNotFound) {
-				locking.Unlock(job.RunnerName, false)
-				return fmt.Errorf("updating runner %s: %w", job.RunnerName, err)
+			if errors.Is(err, runnerErrors.ErrNotFound) {
+				locking.Unlock(job.RunnerName, true)
+				continue
 			}
+			if errors.Is(err, runnerErrors.ErrBadRequest) {
+				// The instance is no longer in a state that can transition to
+				// pending_delete (e.g. it was already deleted or is being deleted
+				// by some other codepath, such as the reconciler noticing it went
+				// missing from the provider/github). There is nothing further for
+				// us to do here. Treating this as fatal would cause this message
+				// to be retried forever, since we never advance past it or delete
+				// it, blocking every subsequent message for this scale set.
+				slog.InfoContext(w.ctx, "runner already being handled/removed; ignoring completed job status update",
+					"runner_name", job.RunnerName, "error", err)
+				locking.Unlock(job.RunnerName, true)
+				continue
+			}
+			locking.Unlock(job.RunnerName, false)
+			return fmt.Errorf("updating runner %s: %w", job.RunnerName, err)
 		}
 		locking.Unlock(job.RunnerName, false)
 	}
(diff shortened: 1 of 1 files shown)


workers/scaleset/scaleset_helper.go
```diff
@@ -160,6 +175,16 @@ func (w *Worker) HandleJobsStarted(jobs []params.ScaleSetJobMessage) (err error)
 				locking.Unlock(job.RunnerName, true)
 				continue
 			}
+			if errors.Is(err, runnerErrors.ErrBadRequest) {
```
gabriel-samfira, line 178: This is interesting. The `HandleJobsStarted()` function is called as a result of a message that comes over the scaleset longpoll session and only happens when a runner transitions from `idle` to `active`. While the runner is `active`, the github API does not allow the runner to be removed.

A reap operation usually tries to remove the runner from github first, then marks the runner as `pending_delete`. The runners should then be cleaned up from the provider. Unless I messed up and didn't do that, there should be no way a runner is reaped by GARM while `active`. If a runner is forcefully removed from the provider while idle, it goes `offline` and cannot pick up a job. So it can't transition to `active` and a message that triggers `HandleJobsStarted()` won't arrive.

If a runner is forcefully removed from the provider while `active` it times out after 10 minutes, but then a `completed` status is sent and a different function should fire.

So I am unsure how this code path was triggered in your env. It may be a bug somewhere else, and I would like to track it down and fix the cause rather than treat the symptom.

Could you share some debug logs by any chance that tracks a runner that has hit this? Are you using agent mode? Were any custom changes made to GARM that could impact this?

workers/scaleset/scaleset_helper.go
```diff
@@ -122,10 +122,25 @@ func (w *Worker) HandleJobsCompleted(jobs []params.ScaleSetJobMessage) (err erro
...
 		_, err := w.store.UpdateInstance(w.ctx, job.RunnerName, runnerUpdateParams)
 		if err != nil {
-			if !errors.Is(err, runnerErrors.ErrNotFound) {
-				locking.Unlock(job.RunnerName, false)
-				return fmt.Errorf("updating runner %s: %w", job.RunnerName, err)
+			if errors.Is(err, runnerErrors.ErrNotFound) {
+				locking.Unlock(job.RunnerName, true)
+				continue
+			}
+			if errors.Is(err, runnerErrors.ErrBadRequest) {
```
gabriel-samfira, line 129: This could indeed race if agent mode is used, but when in agent mode, the runner is set in `pending_delete` and eventually reaped. A retry here would hit `ErrNotFound` and should just continue. But we could find out more from logs if it you could share some (anonymized). You can find me on slack and we can debug this in private if it helps.

gabriel-samfira:
I had a stab at tracking down more of the cause of this issue and opened an alternate PR here:
  * https://github.com/cloudbase/garm/pull/817

Would you mind having a look, and testing it out to see if it fixes your issue?

I also added a deadline for the provider binary exec call. It's [configurable in the provider section](https://github.com/cloudbase/garm/pull/817/changes#diff-39589bfae8cfc756d824142a9c6c6e0a07be11736ff7b7e02cceaa5b27a0274bR60-R63) of the garm config file. It's optional. If not set, GARM will use the pool bootstrap timeout. Whichever is smaller will precede the other.

gabriel-samfira:
I merged the other PR. It should include a fix equal in spirit with what this pr proposes. Give it a shot and let me know if it resolved the issue

gabriel-samfira:
Superseded by #817

PR #823, opened 2026-07-15
provider: fix instance managers wedging on dropped watcher events

benoit-nexthop:
Instance managers acted solely on an in-memory copy of the instance, updated only via watcher events. `consolidateState()` held the manager mutex across provider calls and backoff sleeps (up to minutes), which starved `handleUpdate()` and caused `Update()` to drop events after its 10s send timeout. If the dropped event was the transition back to `pending_delete` after a failed provider delete, the manager was left with a stale "deleting" status - which `consolidateState` had no case for - and idled forever while the runner stayed in `pending_delete` for days.

- `consolidateState` no longer holds the mutex for its duration; it works on a snapshot via new `getInstance()`/`setInstance()` accessors.
- Reconcile against the database on every tick via the new helper `GetInstance()`, so a dropped event can no longer permanently wedge a manager. A missing row exits the manager cleanly.
- Handle `InstanceDeleting` as an interrupted delete and retry it with the normal backoff.

### DIFF
--- a/workers/provider/instance_manager.go
+++ b/workers/provider/instance_manager.go
@@ -271,7 +271,7 @@ func (i *instanceManager) handleCreateInstanceInProvider(instance params.Instanc
 	if err != nil {
 		return fmt.Errorf("updating instance args: %w", err)
 	}
-	i.instance = updated
+	i.setInstance(updated)
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


gabriel-samfira:
Hi @benoit-nexthop 

I will have a look after the end of the month. I am unfortunately in crunch mode until then. Thanks for opening this PR!

benoit-nexthop:
Closing: #871 covers this. The separate consolidation lock and "Resume interrupted deletes in the instance manager" address the wedge this PR worked around, and we have been running main with those changes in production since today without instance managers getting stuck. Thanks!

PR #836, opened 2026-07-30
webapp: add per scale set job queue view

benoit-nexthop:
## What this does

When all runners of a scale set are busy, it is hard to tell where a given workflow job sits in the queue, or why capacity is not being used. This adds a **Job Queue** view to the web UI that groups queued and running jobs per scale set (and per pool, matched by labels), ordered by request time, with links to the GitHub run/job pages and to the scale set/pool detail pages. The view updates live via the existing job/instance websocket events.

## Supporting changes

* **Jobs recorded by scale set listeners now carry the garm scale set ID** (new `WorkflowJob.ScaleSetFkID` column, auto-migrated; exposed as `scale_set_id` on the `Job` API model). Previously the scale set ID was deliberately dropped when recording jobs, so queued jobs could not be attributed to a scale set. Existing rows are backfilled on the next message for that job.
* **The full `RunnerScaleSetStatistic` from each session message is persisted** on the scale set and exposed as `statistics` on the API (previously only `TotalAssignedJobs` survived, as `desired_runner_count`). The view shows GitHub's numbers (assigned jobs, busy/idle runners) next to GARM's instance counts. This makes two things visible that were previously invisible: jobs GitHub holds while a scale set is saturated (GitHub only delivers individual job messages on assignment), and runners GitHub considers offline while GARM thinks they are idle.
* **`consolidateRunnerState` syncs GitHub's per-runner view** (online idle/busy, offline) onto instances that finished installing. Without this, a runner whose agent died after setup stayed "idle" in GARM forever, while GitHub considered it offline and never assigned it jobs — and it counted toward the runner count, starving scale-up. We caught a real production incident (a fleet of registered-but-offline runners with a growing job backlog) within minutes thanks to this divergence being visible in the UI.

## Screenshot

image

## Testing

* `go test -tags testing ./database/... ./workers/...` passes.
* New vitest integration tests for the queue page (grouping, queue ordering, pool label matching, GitHub links, statistics badges).
* Running in our production deployment (org-wide scale sets on a private CloudStack cloud plus a Kubernetes provider).

---

**Update 2026-09-17:** rebased on current main and extended with the follow-ups we have been running in production since:

- webapp: link queued jobs to their GitHub run/job page
- webapp: queue page runner counts from instances, link groups
- scaleset: persist GitHub runner statistics, improve queue view accuracy
- webapp: link runners to their GitHub Actions job

### DIFF
--- a/database/sql/models.go
+++ b/database/sql/models.go
@@ -219,6 +219,10 @@ type ScaleSet struct {
 	Enabled                bool
 	LastMessageID          int64
 	DesiredRunnerCount     int
+	// RunnerStatistics is the last RunnerScaleSetStatistic received from
+	// GitHub on the message session (busy/idle/assigned counts as GitHub
+	// sees them).
+	RunnerStatistics datatypes.JSON
 	// ExtraSpecs is an opaque json that gets sent to the provider
 	// as part of the bootstrap params for instances. It can contain
 	// any kind of data needed by providers.
@@ -473,6 +477,11 @@ type WorkflowJob struct {
 	WorkflowJobID int64 `gorm:"index:workflow_job_id_idx"`
 	// ScaleSetJobID is the job ID for a scaleset job.
 	ScaleSetJobID string `gorm:"index:scaleset_job_id_idx"`
+
+	// ScaleSetFkID is the ID of the scale set that this job was assigned to,
+	// if the job came in through a scale set listener.
+	ScaleSetFkID *uint    `gorm:"index"`
+	ScaleSet     ScaleSet `gorm:"foreignKey:ScaleSetFkID"`
 
 	// RunID is the ID of the workflow run. A run may have multiple jobs.
 	RunID int64
--- a/database/sql/util.go
+++ b/database/sql/util.go
@@ -489,6 +489,14 @@ func (s *sqlDatabase) sqlToCommonScaleSet(scaleSet ScaleSet) (params.ScaleSet, e
 		ret.ProxyName = scaleSet.Proxy.Name
 	}
 
+	if len(scaleSet.RunnerStatistics) > 0 {
+		var stats params.RunnerScaleSetStatistic
+		if err := json.Unmarshal(scaleSet.RunnerStatistics, &stats); err != nil {
+			return params.ScaleSet{}, fmt.Errorf("error unmarshaling runner statistics: %w", err)
+		}
+		ret.Statistics = &stats
+	}
+
 	var ep GithubEndpoint
 	if scaleSet.RepoID != nil {
 		ret.RepoID = scaleSet.RepoID.String()
--- a/webapp/src/lib/api/generated-client.ts
+++ b/webapp/src/lib/api/generated-client.ts
@@ -21,7 +21,9 @@ import {
   ProxiesApi,
   ObjectsApi,
   ToolsApi,
+  JobsApi,
   type GARMAgentRelease,
+  type Job,
   type Repository,
   type Organization,
   type Enterprise,
@@ -152,6 +154,7 @@ export class GeneratedGarmApiClient {
   private proxiesApi: ProxiesApi;
   private objectsApi: ObjectsApi;
   private toolsApi: ToolsApi;
+  private jobsApi: JobsApi;
 
   constructor(baseUrl: string = '') {
     this.baseUrl = baseUrl || window.location.origin;
@@ -188,6 +191,7 @@ export class GeneratedGarmApiClient {
     this.proxiesApi = new ProxiesApi(this.config);
     this.objectsApi = new ObjectsApi(this.config);
     this.toolsApi = new ToolsApi(this.config);
+    this.jobsApi = new JobsApi(this.config);
   }
 
   // Set authentication token
@@ -222,6 +226,7 @@ export class GeneratedGarmApiClient {
     this.providersApi = new ProvidersApi(this.config);
     this.firstRunApi = new FirstRunApi(this.config);
     this.hooksApi = new HooksApi(this.config);
+    this.jobsApi = new JobsApi(this.config);
   }
 
   // Authentication
@@ -642,6 +647,12 @@ export class GeneratedGarmApiClient {
 
   async deleteScaleSet(id: number): Promise<void> {
     await this.scaleSetsApi.deleteScaleSet(id.toString());
+  }
+
+  // Jobs
+  async listJobs(): Promise<Job[]> {
+    const response = await this.jobsApi.listJobs();
+    return response.data || [];
   }
 
   // Instances
--- a/webapp/src/lib/api/generated/api.ts
+++ b/webapp/src/lib/api/generated/api.ts
@@ -2326,6 +2326,12 @@ export interface Job {
      */
     'runner_name'?: string;
     /**
+     * ScaleSetID is the garm ID of the scale set this job was assigned to, if it came in through a scale set listener.
+     * @type {number}
+     * @memberof Job
+     */
+    'scale_set_id'?: number;
+    /**
      * ScaleSetJobID is the job ID when generated for a scale set.
      * @type {string}
      * @memberof Job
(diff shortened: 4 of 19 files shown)


gabriel-samfira:
This looks like an amazing UX improvement!

I will allocate some time to properly review, hopefully next week. Apologies for the delay!

database/sql/models.go
```diff
@@ -474,6 +478,11 @@ type WorkflowJob struct {
 	// ScaleSetJobID is the job ID for a scaleset job.
 	ScaleSetJobID string `gorm:"index:scaleset_job_id_idx"`
 
+	// ScaleSetFkID is the ID of the scale set that this job was assigned to,
+	// if the job came in through a scale set listener.
+	ScaleSetFkID *uint    `gorm:"index"`
+	ScaleSet     ScaleSet `gorm:"foreignKey:ScaleSetFkID"`
```
gabriel-samfira, line 484: I think this needs a:

```
constraint:OnDelete:SET NULL
```

Otherwise we'd get a foreign key constraint error when trying to remove a scaleset if a job is associated with it. We have guards to not allow a scaleset to be removed if it has a runner, but a runner may not be spun up instantly as a job is recorded.

webapp/src/lib/api/generated/api.ts
```diff
@@ -2325,6 +2325,12 @@ export interface Job {
      * @memberof Job
      */
     'runner_name'?: string;
+    /**
```
gabriel-samfira, line 2328: hmm. Did you use:

```
make generate
```

to update this file, or was it manually edited? This file is meant to be generated from the swagger definitions.

Have a look here:

https://openapi-generator.tech/docs/installation/

and here:

https://github.com/cloudbase/garm/blob/main/webapp/DEV_SETUP.md

webapp/src/routes/queue/+page.svelte
```diff
@@ -0,0 +1,408 @@
...
+		if (!runUrl) return '';
+		if (job.workflow_job_id) {
+			return `${runUrl}/job/${job.workflow_job_id}`;
+		}
+		return runUrl;
+	}
+
+	function waitingFor(job: Job): string {
+		if (!job.created_at) return '-';
+		const seconds = Math.max(0, Math.floor((currentTime - new Date(job.created_at).getTime()) / 1000));
```
gabriel-samfira, line 184: This will not trigger on `currentTime`. Bellow in this file you set up a ticker in the `onMount()` function. That ticker updates `currentTime` every 5 seconds, but in the template bellow, `waitingFor()` only accepts a job. So unless a job is refreshed, `waitingFor()` is never fired.

Is that intended? If so, do we still need a ticker for `currentTime`?

If it's not intended, the `waitingFor()` function needs to accept a `now: number` parameter, and we can then call it like:

```
waitingFor(job, currentTime)
```

webapp/src/routes/queue/+page.svelte
```diff
@@ -0,0 +1,408 @@
...
+										</td>
+										<td class="px-4 py-2 text-sm">
+											<div class="flex flex-wrap gap-1">
+												{#each job.labels || [] as label}
+													<span class="inline-flex items-center px-2 py-0.5 rounded text-xs font-medium bg-gray-100 text-gray-800 dark:bg-gray-700 dark:text-gray-300">{label}</span>
+												{/each}
+											</div>
+										</td>
+										<td class="px-4 py-2 text-sm text-gray-500 dark:text-gray-400 whitespace-nowrap">
+											{waitingFor(job)}
```
gabriel-samfira, line 397: If we want this to trigger every 5 seconds, we need to change this to:

```
{waitingFor(job, currentTime)}
```

See above comments.

webapp/src/routes/queue/+page.svelte
```diff
@@ -0,0 +1,408 @@
...
+		// Runners whose agent GitHub reports as offline (won't receive jobs).
+		offlineRunners: number;
+		// Instances GARM is still creating (not yet in running state).
+		provisioning: number;
+		// GitHub's own view of the scale set (busy/idle runners, assigned jobs),
+		// from the last message session statistics. Divergence from the GARM
+		// instance counts usually means runners GitHub considers offline/gone.
+		githubStats?: RunnerScaleSetStatistic;
+		queued: Job[];
+		inProgress: Job[];
```
gabriel-samfira, line 45: I may be missing it, but is `inProgress` ever rendered?

database/sql/util.go
```diff
@@ -489,6 +489,14 @@ func (s *sqlDatabase) sqlToCommonScaleSet(scaleSet ScaleSet) (params.ScaleSet, e
 		ret.ProxyName = scaleSet.Proxy.Name
 	}
 
+	if len(scaleSet.RunnerStatistics) > 0 {
+		var stats params.RunnerScaleSetStatistic
+		if err := json.Unmarshal(scaleSet.RunnerStatistics, &stats); err != nil {
+			return params.ScaleSet{}, fmt.Errorf("error unmarshaling runner statistics: %w", err)
```
gabriel-samfira, line 495: If we ever run into a corrupted stat, listing scalesets will fail. This should probably be logged as an error, but not bail here. See:

https://github.com/cloudbase/garm/blob/2edcc2fff7ccd44343119c06e7e68a79509817eb/database/sql/util.go#L81-L88

as a reference.

webapp/src/lib/api/generated-client.ts
```diff
@@ -644,6 +649,12 @@ export class GeneratedGarmApiClient {
     await this.scaleSetsApi.deleteScaleSet(id.toString());
   }
 
+  // Jobs
+  async listJobs(): Promise<Job[]> {
+    const response = await this.jobsApi.listJobs();
```
gabriel-samfira, line 654: This might have the potential to overload the DB. If the jobs table stays small, this should be fine, but if we ever end up with many jobs, it might impact performance. Worth keeping track of.

webapp/src/routes/queue/+page.svelte
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	function runnerCounts(list: Instance[]) {
+		const running = list.filter((i) => i.status === 'running');
+		const provisioning = list.filter((i) =>
+			['pending_create', 'creating', 'pending'].includes(i.status || '')
```
gabriel-samfira, line 51: The valid instance states are:

https://github.com/cloudbase/garm-provider-common/blob/4f5d9bf8094976b96c9e8db1e4f4155afd846ba9/params/params.go#L43-L54

The state `pending` is not a valid state.

database/sql/models.go
```diff
@@ -219,6 +219,10 @@ type ScaleSet struct {
 	Enabled                bool
 	LastMessageID          int64
 	DesiredRunnerCount     int
+	// RunnerStatistics is the last RunnerScaleSetStatistic received from
+	// GitHub on the message session (busy/idle/assigned counts as GitHub
+	// sees them).
+	RunnerStatistics datatypes.JSON
```
gabriel-samfira, line 225: The model changes will require a migration to be added. We switched to `gormigrate` to do away with the custom code in `sql.go` (and the convoluted conditionals we used to have).

A file in `database/sql/migrations/0007_scaleset_job_queue.go` with something like:

```go
package migrations

import (
	"github.com/go-gormigrate/gormigrate/v2"
	"gorm.io/datatypes"
	"gorm.io/gorm"
)

// scaleSet0007 adds the cached RunnerScaleSetStatistic received from
// GitHub on the message session.
// This is just a stub, AutoMigrate ignores everything else.
type scaleSet0007 struct {
	RunnerStatistics datatypes.JSON
}

func (scaleSet0007) TableName() string {
	return "scale_sets"
}

// workflowJob0007 adds the scale set attribution for jobs recorded by
// scale set listeners.
type workflowJob0007 struct {
	ScaleSetFkID *uint `gorm:"index"`
}

func (workflowJob0007) TableName() string { return "workflow_jobs" }

func init() {
	Register(&gormigrate.Migration{
		ID: "0007_scaleset_job_queue",
		Migrate: func(tx *gorm.DB) error {
			if err := tx.AutoMigrate(&scaleSet0007{}); err != nil {
				return err
			}
			return tx.AutoMigrate(&workflowJob0007{})
		},
	})
}
```

At the end, your migrations table should look like this:

```
sqlite> select * from migrations;
SCHEMA_INIT
0001_baseline
0002_case_insensitive_indexes
0002_lower_indexes
0003_forge_instances
0004_garm_agent_version
0005_allow_insecure_garm_agent
0006_proxies
0007_scaleset_job_queue
sqlite> 
```

gabriel-samfira (changes requested):
Just a few comments.

benoit-nexthop:
Rebased on current main (61976c1b) and added the follow-up commits listed in the description. Builds, tests and lint pass on the branch.

gabriel-samfira:
would you mind having a look at the comments I left on the PR?

PR #841, opened 2026-08-06
scaleset: reap runners stranded by GitHub Actions outages

benoit-nexthop:
Stacked on #836 — the first commit here is that PR's status-sync change, which this depends on (it is what surfaces GitHub's offline/idle/busy view onto instances during consolidation). Once #836 merges and this branch is rebased, the diff collapses to the single reaper commit.

During the 2026-08-06 GitHub Actions outage, hundreds of scale set runners ended up stranded in **running/offline**: the agent registered, then died or lost its connection to the actions service. GARM only tracked GitHub's offline status as metadata and never recycled such runners, pinning every scale set at max_runners and starving the fleet.

Changes:

- Track when each runner is first observed in running/offline (in memory on the scale set worker; there is no status-changed timestamp in the DB and UpdatedAt is bumped by every write) and recycle it after `DefaultRunnerOfflineTimeout` (10 minutes) plus a deterministic per-runner jitter (fnv hash of the name, up to 5 minutes). The jitter prevents runners that went offline together from being reaped, respawned and going offline together in ever more synchronized batches. Offline status is only refreshed by the periodic consolidation pass, so effective reap latency is 10-15+ minutes.
- Cap offline reaps at `max(5, max_runners/4)` per consolidation pass. Provider delete/create operations fan out one goroutine per instance, and unbounded churn can overwhelm the provider API.
- When GitHub refuses to deregister a runner (`TaskAgentJobStillRunningException`, seen when the broker lost a job assignment the runner never acquired), log the refusal and proceed with pending_delete instead of aborting. Destroying the instance is exactly what un-sticks GitHub's state: the agent disappears, GitHub fails the stuck job and releases the runner.

Caveat learned in production: the GitHub runners list is **not** a reliable busy indicator for scale set runners — it reports idle for runners that are actively executing a job. Runners GARM considers active are therefore never touched based on the list (an earlier iteration of this patch did, and cancelled running jobs).

With this, scale sets keep cycling stranded runners during an outage so fresh runners can pick up whatever jobs do go through, and the fleet recovers on its own once GitHub stabilizes. Battle-tested in production during the outage itself: ~340 stranded runners across 5 scale sets were recycled automatically and the fleet tracked GitHub's recovery with no human intervention.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/workers/scaleset/scaleset.go
+++ b/workers/scaleset/scaleset.go
@@ -457,7 +457,8 @@ func (w *Worker) consolidateRunnerState(runners []params.RunnerReference) error 
 	// Cross check what exists in github with what we have in the database.
 	for name, runner := range ghRunnersByName {
 		status := runner.GetStatus()
-		if _, ok := dbRunnersByName[name]; !ok {
+		dbRunner, ok := dbRunnersByName[name]
+		if !ok {
 			// runner appears to be active. Is it not managed by GARM?
 			if status != params.RunnerIdle && status != params.RunnerActive {
 				slog.InfoContext(w.ctx, "runner does not exist in GARM; removing from github", "runner_name", name)
@@ -470,6 +471,39 @@ func (w *Worker) consolidateRunnerState(runners []params.RunnerReference) error 
 			}
 			continue
 		}
+
+		// Sync github's view of the runner (online idle/busy or offline) onto
+		// the instance. Without this, a runner whose agent died after setup
+		// stays "idle" in GARM forever, while github considers it offline and
+		// never assigns it jobs. Only touch runners that finished installing;
+		// runners in earlier lifecycle states are expected to be offline.
+		switch dbRunner.RunnerStatus {
+		case params.RunnerIdle, params.RunnerActive, params.RunnerOffline:
+		default:
+			continue
+		}
+		switch status {
+		case params.RunnerIdle, params.RunnerActive, params.RunnerOffline:
+		default:
+			continue
+		}
+		if dbRunner.RunnerStatus == status {
+			continue
+		}
+		if ok := locking.TryLock(name, w.consumerID); !ok {
+			slog.DebugContext(w.ctx, "runner is locked; skipping runner status sync", "runner_name", name)
+			continue
+		}
+		slog.InfoContext(w.ctx, "syncing runner status from github", "runner_name", name, "old_status", dbRunner.RunnerStatus, "new_status", status)
+		updatedRunner, err := w.store.UpdateInstance(w.ctx, name, params.UpdateInstanceParams{RunnerStatus: status})
+		locking.Unlock(name, false)
+		if err != nil {
+			if !errors.Is(err, runnerErrors.ErrNotFound) {
+				slog.ErrorContext(w.ctx, "error updating runner status", "runner_name", name, "error", err)
+			}
+			continue
+		}
+		w.runners[updatedRunner.ID] = updatedRunner
 	}
 
 	unlockFn, err := w.reapTimedOutRunners(ghRunnersByName)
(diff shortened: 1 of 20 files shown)


benoit-nexthop:
Rebased on current main (61976c1b). This no longer stacks on #836 and merges independently. The reaper was reconciled with the offline-runner handling that landed upstream since (it records the new `offline_timeout` lifecycle outcome), and includes a follow-up fix so runners the GitHub list reports as idle are not reaped while a job is being assigned. Builds, tests and lint pass.

workers/scaleset/scaleset.go
```diff
@@ -405,7 +416,13 @@ func (w *Worker) removeRunnerFromGithubAndSetPendingDelete(runnerName string, ag
 	}
 	if err := scaleSetCli.RemoveRunner(w.ctx, agentID); err != nil {
 		if !errors.Is(err, runnerErrors.ErrNotFound) {
-			return fmt.Errorf("removing runner %s: %w", runnerName, err)
+			// Github may refuse to deregister a runner it believes is running a
```
gabriel-samfira: The whole point of this is to gate against accidental removal of runners. Removing the VM in this case will not unstick anything in github. It removes a runner which is potentially running a job, after a 10 minute timeout GitHub cancels the job, which auto removes the runner from GitHub and sends the `completed` event.

But this will break healthy runners that we mistakenly see as not there, which is undesirable.

workers/scaleset/scaleset.go
```diff
@@ -456,8 +473,34 @@ func (w *Worker) recordLifecycleEvent(outcome string) {
...
+	h := fnv.New64a()
+	h.Write([]byte(runnerName))
+	return time.Duration(h.Sum64() % uint64(appdefaults.DefaultRunnerOfflineTimeout/2))
+}
+
+// maxOfflineReapsPerPass bounds how many offline runners we recycle in one
+// consolidation pass, so a mass-offline event (github outage) trickles
+// delete/create calls to the provider over several passes instead of issuing
+// them all at once.
+func (w *Worker) maxOfflineReapsPerPass() int {
```
gabriel-samfira: Throttling scale down is a good idea. Kind of mirrors the scale-down logic in pools, but in pools we halve (rounding up) at every tick. So you go from 10 to 5, then 2 then 1, then 0.

We also have a cleanup throttle in pools as well separate from the scale-down I mentioned. Same use case as here. Prevent GARM from draining API rate limits and killing resources because of a cleanup event.

workers/scaleset/scaleset.go
```diff
@@ -456,8 +473,34 @@ func (w *Worker) recordLifecycleEvent(outcome string) {
...
 }
 
+// offlineTimeoutJitter returns a deterministic per-runner offset in
+// [0, DefaultRunnerOfflineTimeout/2) added to the offline timeout. Runners
+// that go offline together (e.g. during a github outage) would otherwise be
+// reaped together, respawned together and go offline together again, with the
+// batches growing more synchronized every cycle and hammering the provider
+// API. Hashing the name spreads each batch over a jitter window without
+// keeping extra state.
+func offlineTimeoutJitter(runnerName string) time.Duration {
```
gabriel-samfira: This can be replaced with:

```go
rand.Int(rand.Reader, big.NewInt(appdefaults.DefaultRunnerOfflineTimeout/2))
```

see:

https://github.com/cloudbase/garm/blob/0d35fe0db03b0fcc093976bc231d74e1ee9e57c6/workers/cache/tool_cache.go#L171
https://github.com/cloudbase/garm/blob/0d35fe0db03b0fcc093976bc231d74e1ee9e57c6/util/github/scalesets/message_sessions.go#L158

Issue #847, opened 2026-08-16
In-memory instance cache leaks deleted instances, causing runner scale-up starvation


benoit-nexthop:
## Summary

The in-memory instance cache (`cache/instance_cache.go`) accumulates stale entries for instances that were long ago deleted from the database. Over ~9 days of normal operation our GARM instance's cache grew to ~900 instances while the database contained only 77, of which 819 cached entries were tombstones with `status: deleted`. Anything served from the cache (e.g. the `/api/v1/ws/metrics` dashboard snapshot, per-pool/scale-set runner counts, etc.) reports wildly inflated numbers. Restarting GARM "fixes" it (cache reloads from the DB), and the leak then starts again.

## Environment

- GARM main (reproduced against current `main`, commit `2edcc2ff`)
- 6 scale sets (~60–80 concurrent runners, frequent scale up/down churn)

## Observed behavior

Dashboard reported:

```
Instances: 895   (16 active / 329 idle / 53 pending / 497 offline)
```

while `GET /api/v1/instances` returned 77 instances, all `running`. Dumping a snapshot from `/api/v1/ws/metrics` shows the cache contents:

```
total instances in snapshot: 896
by instance status: deleted: 819, running: 74, deleting: 2, pending_delete: 1
by runner status:   terminated: 494, idle: 328, active: 16, pending: 48, installing: 5, offline: 5
```

819 cache entries have `status: deleted` — the terminal status set just before the DB row is removed — and they are never evicted.

## Root cause

Instance deletion emits two watcher events in sequence:

1. `UpdateOperation` with `Status = InstanceDeleted` (set by the provider worker via `SetInstanceStatus`)
2. `DeleteOperation` when the row is removed (`DeleteInstanceByName` / `DeleteInstance`)

The cache worker (`workers/cache/cache.go`, `handleInstanceEvent`) stores the instance on create/update and evicts it on delete. This relies on events being delivered **in order and reliably**. The watcher guarantees neither:

### 1. Events are delivered out of order

`database/watcher/watcher.go`, `serviceProducer`:

```go
case payload := <-prod.messages:
    w.mux.Lock()
    for _, c := range w.consumers {
        go c.Send(payload)   // one goroutine per event per consumer
    }
    w.mux.Unlock()
```

[...]

PR #850, opened 2026-08-17
watcher: deliver events to consumers in order, without drops

benoit-nexthop:
Fixes the event delivery guarantees of the database watcher. Addresses the root cause of #847.

The watcher currently fans out each event to every consumer in a fresh goroutine (`go c.Send(payload)`), so two events emitted back to back can be delivered in reverse order. Instance deletion emits `update(status=deleted)` followed by delete. When a consumer observes them in the wrong order, the late update re-inserts an already deleted instance into event-driven state. Since the status is terminal, nothing ever evicts it, it'll linger forever. In our production deployment this leaked hundreds of phantom instances over 9 days (dashboard reported 895 instances vs 77 real) and starved scale-up, with queued jobs waiting up to 17 minutes for a runner on an otherwise idle system.

Consumer channels are also buffered at 1 with a 1 second send timeout that silently drops events during bursts, logged only at debug level.

This PR replaces the per-event goroutine with a synchronous enqueue into an ordered per-consumer queue, drained by a dedicated dispatch goroutine:
- Events are delivered to each consumer in emission order.
- Send never blocks on a slow consumer and never drops events; a warning is logged if a consumer's queue keeps growing.
- The dispatch loop owns closing the messages channel, so a send on a closed channel cannot happen.

Includes regression tests for in-order delivery and for no drops with a slow consumer under burst.

This is only the first step of fixing #847: it removes the source of the corruption, but event-driven state should also self-heal if an event is ever missed (e.g. periodic reconciliation of the instance cache and scale set worker state against the database), and consumers should evict on `update(status=deleted)` as defense in depth. I'll send follow up changes once this one lands.

### DIFF
--- a/database/watcher/watcher.go
+++ b/database/watcher/watcher.go
@@ -119,8 +119,12 @@ func (w *watcher) serviceProducer(prod *producer) {
 			return
 		case payload := <-prod.messages:
 			w.mux.Lock()
+			// Send synchronously and in registration-independent but stable
+			// per-consumer order. Send only enqueues (it cannot block on a slow
+			// consumer), and calling it inline preserves event ordering for each
+			// consumer.
 			for _, c := range w.consumers {
-				go c.Send(payload)
+				c.Send(payload)
 			}
 			w.mux.Unlock()
 		}
@@ -140,7 +144,9 @@ func (w *watcher) RegisterConsumer(ctx context.Context, id string, filters ...co
 		id:       id,
 		ctx:      ctx,
 	}
+	c.cond = sync.NewCond(&c.mux)
 	w.consumers[id] = c
+	go c.dispatch()
 	go w.serviceConsumer(c)
 	return c, nil
 }
--- a/database/watcher/consumer.go
+++ b/database/watcher/consumer.go
@@ -18,20 +18,26 @@ import (
 	"context"
 	"log/slog"
 	"sync"
-	"time"
 
 	"github.com/cloudbase/garm/database/common"
 )
+
+// queueWarnThreshold is the pending-queue depth at which we start warning
+// that a consumer is not keeping up with the event stream. Events are never
+// dropped; this only makes a slow consumer visible.
+const queueWarnThreshold = 512
 
 type consumer struct {
 	messages chan common.ChangePayload
 	filters  []common.PayloadFilterFunc
 	id       string
 
-	mux    sync.Mutex
-	closed bool
-	quit   chan struct{}
-	ctx    context.Context
+	mux     sync.Mutex
+	cond    *sync.Cond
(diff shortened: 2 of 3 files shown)


database/watcher/watcher.go
```diff
@@ -119,8 +119,12 @@ func (w *watcher) serviceProducer(prod *producer) {
 			return
 		case payload := <-prod.messages:
 			w.mux.Lock()
+			// Send synchronously and in registration-independent but stable
+			// per-consumer order. Send only enqueues (it cannot block on a slow
+			// consumer), and calling it inline preserves event ordering for each
+			// consumer.
 			for _, c := range w.consumers {
-				go c.Send(payload)
+				c.Send(payload)
```
gabriel-samfira, line 127: I was actually looking at this bit of code. You can use a WaitGroup{} here to send notifications to all consumers in parallel. Events would still be sent synchronously, but if we have multiple slow consumers, the wait time would not compound. We would wait at most 1 second, regardless of how many slow consumers we have.

Something like:

```go
wg := sync.WaitGroup{}
for _, c := range w.consumers {
    wg.Go(func(){
        c.Send(payload)
    })
}
wg.Wait()
```

gabriel-samfira:
Thanks for this PR. I think the fix for the issue you've been seeing will need fixes in multiple places. I think the pain we've been having with syncing state and acting on runner transitions (from pending_create to creating to running to deleting, etc) is due to a more fundamental flaw in how we track state.

There is a desired state and an observed state that I sadly mashed together. The desired state is what the user expresses and what the code tries to converge to. The observed state is the state of the resource we need (the runner in our case) as it's being created/updated/removed by a provider. I need to think of how we can separate the two. That way, when the provider acts on a snapshot of the desired state (a runner was placed in pending create), we can still issue a `force_delete` on the runner without interfering with the provider. The provider doesn't need to lock the whole instance as it's creating it. Once its done and it sees that the new desired state is to remove it, it can just act again.

This change is still good, and I will have a closer look at it once I get back from vacation, but I also need to have a closer look at how we handle state, so we don't just keep plugging holes here and there.

gabriel-samfira:
Would you mind giving #871 a try to see if it fixes the issue you observed? There were several other issues that could have contributed to the same issue that I coalesced in that PR.

benoit-nexthop:
Closing: the watcher changes in #871 ("Watcher bugfixes") wait for every consumer per message and add a per-consumer dispatch goroutine with deeper buffers, which fixes the reordering this PR targeted. We upgraded our production deployment to current main today and the events are delivered in order. I'll send the regression tests plus the periodic reconciliation / defense-in-depth changes mentioned above as a separate PR.

PR #852, opened 2026-08-19
Retry failed gitea tools updates

katexochen:
The giteaUpdateLoop skipped the update whenever metadataURL or useInternal differed from the last successful update, even if giteaToolsUpdateDeadline passed. When the initial fetch failed, this resulted in the tools never being fetched, as oldMetadataURL/oldUseInternal are only assigned after the first successful fetch.

The condition missed a proper application of De Morgan's law. The condition is right further up in the function where it isn't negated. The intended logic is "fetch when stale OR the URL changed OR the internal flag flipped".

### DIFF
--- a/workers/cache/tool_cache.go
+++ b/workers/cache/tool_cache.go
@@ -224,7 +224,7 @@ reset:
 			return
 		case <-ticker.C:
 			now := time.Now().UTC()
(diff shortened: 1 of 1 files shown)


gabriel-samfira (approved):
Thank you!

PR #853, opened 2026-08-19
Add --password-file to init and profile commands

katexochen:
Passing the admin password with --password puts it in the process arguments, where it is visible to other users on the machine and ends up in logs of whatever runs the command. Add --password-file to read it from a file instead, which allows automating the initialization without exposing the password. profile add and profile login take the same flag, as they have the same problem.

### DIFF
--- a/cmd/garm-cli/cmd/init.go
+++ b/cmd/garm-cli/cmd/init.go
@@ -67,6 +67,10 @@ garm-cli init --name=dev --url=https://runner.example.com --username=admin --pas
 				return fmt.Errorf("a manager with name %s already exists in your local config", loginProfileName)
 			}
 		}
+		if err := resolvePasswordFile(); err != nil {
+			return err
+		}
+
 		var caBundle []byte
 		var err error
 		if controllerCABundle != "" {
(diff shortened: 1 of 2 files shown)


katexochen:
There are other secrets I'd like to set via file, too. Happy to follow up with that if desired.

gabriel-samfira:
I've been thinking about using environment variables to pass in secrets to the `init` command. This would allow container deployments to more easily initialize a controller on first start. Would that be easier than using a file?

They're not mutually exclusive. We could have both, with one of them being authoritative. What do you think?

gabriel-samfira:
Also, rather than having a separate file per setting, we could add a `--config-file`  flag that allows the user to set structured settings in toml/yaml/json (not sure which would be best, no strong opinions on my part - bonus if there is a library that could handle any). We could then merge settings using something like:

* `--config-file` sets defaults (or env variables - we can just pick one to start with, no need to conflate a single PR)
* explicit flags override defaults
* any required missing flags are prompted interactively

something like `garm-cli init --config-file ./garm.[toml|yaml] --username admin`. In this case, even if we have a username in the config file, it would be overriden by the explicit `--username` flag.

This is already somewhat done for missing command line args.

katexochen:
I want to use this with systemd credentials, which are delivered via `LoadCredential` as file in `$CREDENTIALS_DIRECTORY`, so a `--password-file` flag is a direct hand over. An env var would need a wrapper to read the file and export it. Same goes for a config file, which would make it even more complicated to construct on the fly.

gabriel-samfira:
ahh. Ok. Now it makes sense.

gabriel-samfira (approved):

PR #855, opened 2026-08-20
Fix runner records stranded in pending_delete wedging scale set autoscaling

### LINKED ISSUE #854: Runner records stranded in pending_delete permanently wedge scale set autoscaling
### Summary

A runner record can get stranded in `pending_delete` forever while its provider instance is already gone. Stranded records count toward the scale set's runner count, so once a scale set has `stuck records >= min(desired, max_runners)` its autoscaler sits in the scale-down branch permanently and never creates a runner again. Jobs then queue indefinitely on a scale set that is actually empty. With `min_idle_runners: 0`, `desired` is typically 1 on a quiet set, so two stuck records are enough to wedge any scale set regardless of `max_runners`. A controller restart clears it (state is rebuilt from the DB), which is how we kept recovering before finding the cause.

Observed repeatedly in a production deployment built from main at `afda4e76f1808e8b41d72edd0c17f99a51d84758` with the GCP provider (v0.1.5) and agent mode enabled. Under load we saw multiple new stranded records per day.

### The chain, verified in source and controller logs

1. A job completes; the scaleset worker moves the instance to `pending_delete` (validated write). The instance manager receives the update and acts: `consolidateState()` force-writes `deleting`, then calls the provider delete while holding `i.mux`. A GCE delete takes tens of seconds.
2. The watcher event for that `deleting` write is received by the manager's `updatesLoop`, which immediately blocks in `handleUpdate()` on `i.mux`. Every later event for the same instance is then dropped by `instanceManager.Update()`'s 10 s send timeout:

   ```
   01:05:20.192 INFO  job completed
   01:05:20.627 INFO  deleting instance in provider
   01:05:30.669 ERROR failed to update instance | timeout while sending update to instance manager
   01:05:40.669 ERROR failed to update instance | timeout while sending update to instance manager
   ```

3. Meanwhile the agent worker sees the runner report terminated (`handleStatusMessage`) and calls `Runner.SetInstanceToPendingDelete`, which uses `ForceUpdateInstance`: it writes `pending_delete` over `deleting` with no transition validation, moving the record backwards mid-delete.
4. The provider delete finishes; the instance is gone. The manager's validated `deleted` write now fails, because `pending_delete -> deleted` is not in `params.InstanceStatusTransitions`:

   ```
   01:05:55.550 ERROR consolidating state | setting instance status to deleted: ... invalid instance status transition from pending_delete to deleted
   ```

5. The blocked `handleUpdate` from step 2 acquires the mutex and sets the manager's cached state to `deleting`. `consolidateState()`'s switch has no case for `InstanceDeleting`, so every 5 s tick silently does nothing, forever. The manager is alive, so `retryFailedRunners` skips it; the scaleset worker's consolidation skips deletion-lane records because the provider worker owns them. Nothing ever re-drives the record.
6. `runnerCount()` is `len(w.runners)` and counts the stranded record. `targetRunners()` is `min(MinIdleRunners+desired, MaxRunners)`, so `handleAutoScale` picks `handleScaleDown` forever, logging `considering runners for removal` every 5 s (we counted 240 lines per 10 minutes during one wedge).

### Remediation that works on a live wedge

`garm-cli runner delete -f <name>` moves the record `pending_delete -> pending_force_delete -> deleting -> deleted` and the set resumes scaling within seconds (verified on two live occurrences). A controller restart also clears it.

### Fixes

I have a PR ready with three independent fixes, each with tests; any one of them breaks the chain, and together they remove the corruption, the dead state, and the starvation:

1. `Runner.SetInstanceToPendingDelete` uses a validated update and treats a refused transition from a deletion-lane status as success (the intent, the runner goes away, is already being fulfilled).
2. `consolidateState()` treats `InstanceDeleting` as a resumable delete (deletes are idempotent: providers report a missing instance as success), requeues on provider errors for every non-forced path, and force-writes the final `deleted` when the row regressed onto the deletion lane behind its back, because the provider resource is confirmed gone at that point.
3. `runnerCount()` excludes deletion-lane records, and `handleScaleDown`'s deletion-lane case no longer consumes the removal delta.

Two related defects the PR deliberately does not touch, flagging them here instead: a timed-out `instanceManager.Update()` send discards the update with no requeue, and `consolidateState()` holds `i.mux` across provider calls, which is what starves `updatesLoop`/`handleUpdate` and causes those drops in the first place. Both are worth fixing but are invasive to the worker's concurrency structure; with the three fixes above the dropped updates become harmless.

KT-Doan:
Closed 2026-09-16: superseded by #871, which fixes this on `main`. Verified against the end-to-end regression from this PR: a wedged scale set converges on `main`, so the change here is no longer needed.

Fixes #854.

Runner records could get stranded in `pending_delete` forever while the provider instance was already gone; stranded records count toward the scale set's runner count, so a handful of them pin the autoscaler in permanent scale-down and jobs queue on an empty scale set. The full verified chain (source references plus controller log excerpts) is in the issue.

Three independent commits, each with tests; any one of them breaks the chain, and together they remove the corruption, the dead state, and the starvation:

1. `runner`: `SetInstanceToPendingDelete` (the agent terminated path) used `ForceUpdateInstance`, which can move a record backwards from `deleting` to `pending_delete` mid-delete, behind the provider worker's back. It now uses a validated update and treats a rejected transition from a deletion-lane status as success, since the runner is already on its way out.

2. `workers/provider`: `consolidateState` had no case for `InstanceDeleting`, so a manager whose cached state landed there (a status update delivered late, after a long provider call, or a restart mid-delete) ticked forever doing nothing. `InstanceDeleting` is now a resumable delete (providers report a missing instance as success, so re-driving is safe), provider errors requeue for every non-forced path instead of falling through to `deleted`, and a `deleted` write refused because the row regressed onto the deletion lane is retried with force, because the provider resource is confirmed gone at that point.

3. `workers/scaleset`: `runnerCount()` counted deletion-lane records, so stuck records suppressed scale-up regardless of `max_runners`, and with `min_idle_runners: 0` two of them wedge a quiet scale set permanently. It now counts live records only, and `handleScaleDown`'s deletion-lane case no longer consumes the removal delta those records are no longer part of.

Testing: new unit tests in `runner` (real sqlite store), `workers/provider` (fake helper plus the generated provider mock), and `workers/scaleset`; the new tests fail on main exactly where the incident behavior lives. Full suite green with the changes (`go test -tags testing ./...`). The remediation path for an already-wedged deployment (`garm-cli runner delete -f`) was verified on two live occurrences before writing these fixes.

Not addressed here, flagged at the end of #854: `instanceManager.Update()` discards a timed-out send with no requeue, and `consolidateState` holds `i.mux` across provider calls, which is what starves `handleUpdate` into those timeouts. Both are worth their own change; with this PR the dropped updates become harmless.

### DIFF
--- a/runner/agent.go
+++ b/runner/agent.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"time"
@@ -9,6 +10,7 @@ import (
 	runnerErrors "github.com/cloudbase/garm-provider-common/errors"
 	commonParams "github.com/cloudbase/garm-provider-common/params"
 	"github.com/cloudbase/garm/auth"
+	garmErrors "github.com/cloudbase/garm/internal/errors"
 	"github.com/cloudbase/garm/params"
 )
 
@@ -60,7 +62,12 @@ func (r *Runner) SetInstanceToPendingDelete(ctx context.Context) error {
 		Status: commonParams.InstancePendingDelete,
 	}
 
-	if _, err := r.store.ForceUpdateInstance(r.ctx, instance.ID, updateParams); err != nil {
+	if _, err := r.store.UpdateInstance(r.ctx, instance.ID, updateParams); err != nil {
+		var te *runnerErrors.InstanceTransitionError
+		if errors.As(err, &te) && garmErrors.InstanceIsBeingDeleted(te.From) {
+			// Already on the deletion lane; treat the refusal as success.
+			return nil
+		}
 		return fmt.Errorf("failed to set instance to pending_delete: %w", err)
 	}
 	return nil
--- a/runner/agent_test.go
+++ b/runner/agent_test.go
@@ -1,0 +1,144 @@
+// Copyright 2025 Cloudbase Solutions SRL
+//
+//    Licensed under the Apache License, Version 2.0 (the "License"); you may
+//    not use this file except in compliance with the License. You may obtain
+//    a copy of the License at
+//
+//         http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+//    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+//    License for the specific language governing permissions and limitations
+//    under the License.
+
+package runner
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	commonParams "github.com/cloudbase/garm-provider-common/params"
+	"github.com/cloudbase/garm/auth"
+	"github.com/cloudbase/garm/database"
+	dbCommon "github.com/cloudbase/garm/database/common"
+	garmTesting "github.com/cloudbase/garm/internal/testing"
+	"github.com/cloudbase/garm/params"
+)
+
+type AgentTestSuite struct {
+	suite.Suite
+
+	adminCtx context.Context
+	store    dbCommon.Store
+	runner   *Runner
+	instance params.Instance
+}
+
+func (s *AgentTestSuite) SetupTest() {
+	dbCfg := garmTesting.GetTestSqliteDBConfig(s.T())
+	db, err := database.NewDatabase(context.Background(), dbCfg)
+	if err != nil {
+		s.FailNow(fmt.Sprintf("failed to create db connection: %s", err))
+	}
+	s.store = db
+
+	s.adminCtx = garmTesting.ImpersonateAdminContext(context.Background(), db, s.T())
+
+	githubEndpoint := garmTesting.CreateDefaultGithubEndpoint(s.adminCtx, db, s.T())
+	testCreds := garmTesting.CreateTestGithubCredentials(s.adminCtx, "test-creds", db, s.T(), githubEndpoint)
+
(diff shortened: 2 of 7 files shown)


KT-Doan:
Added an end-to-end regression test that replays the full interleave from #854 through a real store, the watcher, and a running provider worker (slow provider delete, deleting event parked on the manager mutex, forced pending_delete regressing the row mid-delete, follow-up event dropped by the 10s send timeout). On main the test times out with the record stuck in pending_delete; with this series it converges to deleted within a couple of ticks.

gabriel-samfira:
Hi @KT-Doan 

We need to fix the root cause of this, rather than work around it. It's clear that there is a flaw in the way I designed state handling here, that is easily triggered under load. I am on vacation for a the next couple of weeks, but once I get back this will be first on the list of priorities. We need to iron this out and formalize a better way to keep track of state across workers. Otherwise we'll keep plugging holes as they sprout.

This PR and the ones bellow do a great job at explaining what's wrong. The long term solution may be making a step back and redesigning some things.

I will be away until the 31st of the month. After that I'll be back at it. Apologies for the delay on this.

related:
  * https://github.com/cloudbase/garm/pull/823
  * https://github.com/cloudbase/garm/pull/850

KT-Doan:
Hi @gabriel-samfira 

No worries! This PR is just the implementation we did in our organization and its working pretty smoothly so far, maybe it will help you in the future! Thank you for taking the time to look at this.

gabriel-samfira:
It definitely helps. It demonstrates the failure mode. The other 2 PRs I linked to add more data points. I will try to tackle this as soon as I get back into the office.

gabriel-samfira:
Would you mind trying out #871 to see if it fixes the issue you observed?

PR #866, opened 2026-09-04
Migrate from a concatenated pseudo pool ID

gabriel-samfira:
This change migrates from a pseudo pool ID composed of the entity ID and the name of the scale set, to a stable ID computed from the entity ID and the internal scaleset ID as an UUID V5.

Both the entity ID and the internal scaleset ID are immutable. The UUID V5 will always be 36 characters long, regardless of the input data used to compute it.

The downside of this change is that we now need to do two ListInstance() provider calls to list both instances tagged with the old shape of the ID and the new shape. Internally we record whether or not ListInstances() returned an empty list. If empty, we don't have instances in the provider for a scale set, so on each subsequent list, we only do one call.

So in theory at least, we should only be abusing the IaaS API only as long as there are still old runners tagged with the old shape.

Note: once you migrate to a version that contains this commit, you will no longer be able to migrate back without losing track of runners in the IaaS.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"unicode/utf8"
 
+	"github.com/google/uuid"
 	"github.com/h2non/filetype"
 
 	"github.com/cloudbase/garm-provider-common/cloudconfig"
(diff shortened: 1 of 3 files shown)


cbartz (approved):
LGTM.

Thank you so much for addressing this issue.

We could mention in the PR description that rollback of the GARM version will not be safe (when new runners are created with the new release and folks switch back to the one before), so we can mention this in the release notes. 

Once this is merged, I'll test it and we can close https://github.com/cloudbase/garm-provider-openstack/issues/34 afterwards.

gabriel-samfira:
@cbartz thanks! Description was updated and merged. Feel free to test and let me know if it works for you.

Issue #860, opened 2026-08-25
Deleted instance records leak permanently when the DB watcher drops the final status event under burst deletions


KT-Doan:
### Summary

When many scale-set runners are deleted in one burst, the watcher event that carries an instance's final `deleted` status write can be dropped. The scaleset worker removes an instance's DB row only when it receives that event, so every dropped event leaks a permanent row at status `deleted`. The leaked rows survive indefinitely: the delete API refuses records in that status, nothing reconciles them at runtime, and only a controller restart clears them. After a cancelled workflow batch on 2026-08-24 we found 47 such rows, 18 hours old.

This is a sibling of #854. That issue tracks records stranded in `pending_delete` by a dropped update inside the provider worker's instance manager. This one is a second, independent drop point in `database/watcher`, and it produces a different terminal state that the fixes in #855 do not cover: with those fixes applied, deletion-lane records no longer wedge autoscaling, but the leaked `deleted` rows still accumulate.

Observed on a production deployment built from main at `afda4e76f1808e8b41d72edd0c17f99a51d84758` with the three fixes from #855 applied, the GCP provider v0.1.5, and agent mode enabled.

### The chain, verified in source and controller logs

1. A workflow batch is cancelled. About 47 runners across two scale sets go through deletion within seconds of each other. The provider deletes every VM. 31 records reach `runner_status=terminated`; 16 die mid-bootstrap at `runner_status=pending`.

2. Under the burst the provider worker's update channel saturates, the same starvation #854 describes. One instance from the batch, `garm-itlbwyalbr2q`, shows the provider delete being driven twice and then silence, with the send timeouts nearby (times UTC):

   ```
   13:47:22.196 INFO  runner was terminated
   13:47:46.098 INFO  deleting instance in provider
   13:48:43.561 INFO  deleting instance in provider
   13:48:45.168 INFO  updates channel closed
   14:05:45.233 ERROR failed to update instance | timeout while sending update to instance manager
   14:06:54.645 ERROR failed to update instance | timeout while sending update to instance manager
   ```

3. Each instance's final `deleted` write lands in the DB and the watcher fans it out. `database/watcher/consumer.go` gives a consumer 1 second, then silently drops the payload; the only trace is a DEBUG line:

   ```go
   timer := time.NewTimer(1 * time.Second)
   ...
   case <-timer.C:
       slog.DebugContext(w.ctx, "timeout trying to send payload", "payload", payload)
   ```

[...]

Issue #834, opened 2026-07-29
Scaleset runner cancellation issue


Hdom:
Hello, running into an issue with scalesets where tons of runners are being identified as "missing" which is causing GARM to delete the runner even though it has picked up a job.

Timeline:

- 11:21:57 - Runner is pruned because it was not detected as existent in github.

> time=2026-07-29T11:21:57.250Z level=INFO source=/runner/_work/cloudbase-garm/cloudbase-garm/workers/scaleset/scaleset.go:489 msg="runner does not exist in github; removing from provider" runner_name=gssdue4clml-zaict7cxmusy worker=scaleset-worker-gssdue4clml-2 entity=OPG endpoint=github.com

- 11:22:01.107 - Runner picks up a job

> time=2026-07-29T11:22:01.107Z level=INFO source=/runner/_work/cloudbase-garm/cloudbase-garm/workers/scaleset/scaleset_listener.go:166 msg="job started" job_id=23b2785d-8345-5b9e-b726-a0e97794469f job_name="<job name>" runner_name=gssdue4clml-zaict7cxmusy sub_worker=scaleset-listener worker=scaleset-worker-gssdue4clml-2 entity=ORG endpoint=github.com

- 11:22:01.567 - Runner is deleted in kubernetes API

I also found these logs right before GARM started deleting runners, which might be related:

> time=2026-07-29T11:21:56.438Z level=INFO source=/runner/_work/cloudbase-garm/cloudbase-garm/workers/provider/instance_manager.go:426 msg="updates channel closed" worker=instance-worker-gssdue4clmm-pcfkjfe8h9zm

Some thoughts:

When scaleset goes to deregister runner from github it is told by github that the runner is currently running a job, the flow should probably be deregister first then delete in provider? That way github has a chance to tell you its running something.

[...]

PR #871, opened 2026-09-08
Bug and stability fixes

gabriel-samfira:
This PR addresses a bunch of issues in the scaleset workers, HTTP clients, database watcher and a few other places. Overall the changes include:

* Serialize watcher events both during the emission and the consumption of the events. We should never parallelize messages we send or receive from a channel unless they are completely independent from one another.
* The watcher consumers and producers must be non blocking.
* all HTTP clients should have proper timeouts to prevent situations where one HTTP call hangs everything 

These changes (some inspired from some of the PRs) should fix issues recently reported in:

  * https://github.com/cloudbase/garm/pull/855
  * https://github.com/cloudbase/garm/pull/850
  * https://github.com/cloudbase/garm/pull/823
  * https://github.com/cloudbase/garm/issues/860
  * https://github.com/cloudbase/garm/issues/854
  * https://github.com/cloudbase/garm/issues/834

### DIFF
--- a/database/watcher/consumer.go
+++ b/database/watcher/consumer.go
@@ -18,15 +18,28 @@ import (
 	"context"
 	"log/slog"
 	"sync"
-	"time"
 
 	"github.com/cloudbase/garm/database/common"
 )
 
+// queueWarnThreshold is the consumer queue depth at which we start warning
+// that a consumer is not keeping up. Events are never dropped. The queue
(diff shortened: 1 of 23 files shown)


gabriel-samfira:
Ok. I have a bunch of typos in the commit messages. Too lazy to fix them.

PR #878, opened 2026-09-11
Improve metrics

### LINKED ISSUE #865: Extend `garm_runner_operations_total` to include pool
When using garm we had the question "How many jobs were executed on each garm server?". Since all of our garms have their own garm instance, we couldn't do that by provider, therefore I went with `garm_runner_operations_total` and filtered by `CreateInstance` to get an idea on container creation. Since all our providers are named the same, we cannot distinguish by them.

Question would be if you would consider extending `garm_runner_operations_total` by a `pool` label, so we can separate them?
Of course an alternative would be for us to name the providers individually on each host.

gabriel-samfira:
This change adds a number of new metrics, augments some existing metrics and adds sample Grafana dashboards in the contrib folder.

See doc/monitoring.md for a description of the new metrics.

Fixes: #865

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,17 @@ linters:
     - wastedassign
     - whitespace
   settings:
+    goconst:
+      # goconst v1.11 (bundled since golangci-lint v2.13) started scanning
+      # strings inside composite literals (slice/map/struct literals), which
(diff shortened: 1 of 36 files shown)


gabriel-samfira:
CC @cbartz 

I believe you asked about dashboards at one point. This might be useful.

PR #893, opened 2026-09-17
transport: add configurable HTTP/2 health checks for forge connections

benoit-nexthop:
All requests to GitHub (including the scaleset broker long-polls) are multiplexed over a small number of HTTP/2 connections. When such a connection dies silently (no FIN/RST), every in-flight request blocks until the kernel TCP retransmission timeout fires (~15 minutes), stalling all scaleset workers, as observed in our prod cluster.

Add two new options to the `[default]` config section:

```
  http2_read_idle_timeout - when set (> 0), enables HTTP/2 health
    checks: a PING frame is sent on any connection idle for this
    duration and dead connections are closed, failing in-flight
    requests fast.
  http2_ping_timeout - how long to wait for the PING response before
    declaring the connection dead (defaults to the x/net default, 15s).
```

The health checks apply to the transports built for forge credentials as well as the scaleset client, which previously used the default transport. The stray throwaway `http.Client` in `DeleteRunnerScaleSet` now reuses the scaleset client's transport as well.

Defaults are unchanged: with no config set, no health checks are enabled and the scaleset client keeps using the default transport.

## Testing

Running in our production deployment (GitHub.com, scale sets) since today. `go test ./...` passes.


params/http_transport.go
```diff
@@ -0,0 +1,82 @@
...
+// NewHTTPClient returns an *http.Client suitable for API calls to a forge.
+// If HTTP/2 health checks are configured, the client uses a dedicated
+// transport with health checks enabled; otherwise it uses the default
+// transport, preserving previous behavior.
+func NewHTTPClient() *http.Client {
+	readIdleTimeout, _ := getHTTP2HealthChecks()
+	if readIdleTimeout <= 0 {
+		return &http.Client{}
+	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
gabriel-samfira, line 79: Let's not do this. This will patch the singleton for the entire application just for the benefit of the scaleset client. If we need this only for scaleset long-poll, then we need to own the client in the scaleset client itself, build a proper transport and set it on the scaleset http client.

I will do a more in-depth review later. Need to step out for the day.

params/http_transport.go
```diff
@@ -0,0 +1,82 @@
...
+
+package params
+
+import (
+	"log/slog"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/net/http2"
```
gabriel-samfira, line 23: You don't need this to set `PingTimeout` and `ReadTimeout`:

https://pkg.go.dev/net/http#HTTP2Config

I believe setting `SendPingTimeout` and `PingTimeout` will achieve the same result. And it's already in the standard lib.

PR #894, opened 2026-09-17
metrics: add low-cardinality aggregate and per-scaleset runner/job gauges

benoit-nexthop:
Per-instance and per-job metrics churn labels every time a runner cycles, which makes dashboards and alerts expensive once a deployment has 100+ runners turning over. Add aggregate gauges that keep cardinality bounded by status and entity instead of instance identity:

- garm_runner_count{status,runner_status,pool,provider}
- garm_job_count{status,entity_type,entity_name}
- garm_scaleset_status{name,state,entity_type,entity_name,provider}
- garm_scaleset_runner_count{scaleset_name,status,runner_status,provider}
- garm_scaleset_job_count{scaleset_name,status}

The per-scaleset gauges track capacity and queue depth per scale set, which is what you need to decide when to raise max_runners.

Upstream currently only exposes `garm_scaleset_desired_runner_count` at this granularity; the existing per-instance / per-job gauges carry instance names as labels and churn constantly. These gauges are what we drive our Grafana dashboards and alerts from in production.


metrics/scaleset.go
```diff
@@ -19,45 +19,32 @@ import (
 )
 
 var (
-	ScaleSetInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
gabriel-samfira, line 22: Removing this will break existing [dashboards](https://github.com/cloudbase/garm/tree/main/contrib/grafana). Any reason to remove this?

metrics/scaleset.go
```diff
@@ -19,45 +19,32 @@ import (
...
+	// ScaleSetStatus reports the status of each scaleset.
+	// The value is 1 if the scaleset is enabled, 0 if disabled.
 	ScaleSetStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsScaleSetSubsystem,
 		Name:      "status",
-		Help:      "Status of the scale set",
-	}, []string{"id", "enabled", "state"})
-
-	ScaleSetMaxRunners = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
gabriel-samfira, line 36: why remove this? If the goal is to know when you need to raise this, don't you need it?

metrics/scaleset.go
```diff
@@ -19,45 +19,32 @@ import (
...
+	// instance status and runner status. Use this to track per-scaleset capacity
+	// and utilization.
+	ScaleSetRunnerCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsScaleSetSubsystem,
-		Name:      "min_idle_runners",
-		Help:      "Minimum number of idle runners in the scale set",
-	}, []string{"id"})
-
-	ScaleSetDesiredRunnerCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
gabriel-samfira, line 50: Same with this. If the goal is to know when to raise max runners, you will need most of the gauges you're removing here.

metrics/scaleset.go
```diff
@@ -19,45 +19,32 @@ import (
...
-	}, []string{"id", "scaleset_id", "name", "image", "flavor", "prefix", "os_type", "os_arch", "tags", "provider", "runner_group", "scaleset_owner", "scaleset_type"})
-
+	// ScaleSetStatus reports the status of each scaleset.
+	// The value is 1 if the scaleset is enabled, 0 if disabled.
 	ScaleSetStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsScaleSetSubsystem,
 		Name:      "status",
-		Help:      "Status of the scale set",
-	}, []string{"id", "enabled", "state"})
```
gabriel-samfira, line 34: Relabeling/removing these breaks existing dashboards.

metrics/scaleset.go
```diff
@@ -19,45 +19,32 @@ import (
...
-
-	ScaleSetMaxRunners = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Namespace: metricsNamespace,
-		Subsystem: metricsScaleSetSubsystem,
-		Name:      "max_runners",
-		Help:      "Maximum number of runners in the scale set",
-	}, []string{"id"})
+		Help:      "Status of each scaleset (1=enabled, 0=disabled)",
+	}, []string{"name", "state", "entity_type", "entity_name", "provider"})
```
gabriel-samfira, line 29: Scaleset names are mutable (you can rename a scaleset). IDs are not.

runner/metrics/instance.go
```diff
@@ -44,69 +40,160 @@ func CollectInstanceMetric(ctx context.Context, r *runner.Runner) error {
...
+		var info entityInfo
+		var entityID string
+
+		// Look up entity info from pool or scaleset
+		if instance.PoolID != "" {
+			info = poolInfo[instance.PoolID]
+			entityID = instance.PoolID
+		} else if instance.ScaleSetID > 0 {
+			info = scalesetInfo[instance.ScaleSetID]
+			entityID = fmt.Sprintf("scaleset-%d", instance.ScaleSetID)
```
gabriel-samfira, line 136: Why conflate the same field with either a UUIDv4 or a `scaleset-` prefixed string rather than having separate fields for this?

runner/metrics/instance.go
```diff
@@ -44,69 +40,160 @@ func CollectInstanceMetric(ctx context.Context, r *runner.Runner) error {
...
 			string(instance.Status),       // label: status
 			string(instance.RunnerStatus), // label: runner_status
-			parent.OwnerName,              // label: pool_owner
-			parent.Type,                   // label: pool_type
-			poolID,                        // label: pool_id
-			scaleSetID,                    // label: scaleset_id
-			parent.ProviderName,           // label: provider
+			info.Name,                     // label: pool_owner
+			info.Type,                     // label: pool_type
+			entityID,                      // label: pool_id
```
gabriel-samfira, line 150: For scale sets, this will show up as `pool_id="scaleset-<ID>"`, which is an overload that looks architecturally weird :). It also breaks a documented contract. `garm_runner_status` guarantees exactly one of `pool_id`/`scaleset_id` is set per series, and the shipped dashboards rely on `{pool_id!=""}` meaning "pool-owned". If we change this scale-set runners start matching those selectors. `pool_id` needs to stay empty here. `scaleset_id` already carries the identity.

runner/metrics/instance.go
```diff
@@ -44,69 +40,160 @@ func CollectInstanceMetric(ctx context.Context, r *runner.Runner) error {
...
+	type countKey struct {
+		Status       string
+		RunnerStatus string
+		PoolOwner    string
+		PoolType     string
+		Provider     string
+	}
+	counts := make(map[countKey]int)
+
+	// Aggregate counts by scaleset name/status/runner_status/provider
```
gabriel-samfira, line 117: This is nice to have. But Scalesets are not globally unique. Neither are scaleset IDs (github IDs that is, the ones in GARM are unique). So you will have duplicates and you most likely will double count instances on a large enough instance.

For example, if you have 2 repos and each define a `ubuntu-noble` scaleset, that is perfectly valid. Moreover, scaleset names are unique only per entity AND per runner group. For example, in the same entity, you may have an `ubuntu-noble` scaleset in the `Default` runner group and an `ubuntu-noble` scaleset in a `restricted` runner group.

runner/metrics/jobs.go
```diff
@@ -48,5 +61,78 @@ func CollectJobMetric(ctx context.Context, r *runner.Runner) error {
...
+		ScaleSetName string
+		Status       string
+	}
+	scalesetJobCounts := make(map[scalesetJobKey]int)
+
+	for _, job := range jobs {
+		// Determine entity type and name
+		var entityType, entityName string
+		switch {
+		case job.OrgID != nil:
```
gabriel-samfira, line 87: For this, all 3 can be set. Let me explain.

A job is always generated by a workflow on a repo. The webhooks we receive from github will depend on which level we have created the entity and installed a webhook. Conceivably, a user may have an enterprise defined in GARM, as well as an org under that enterprise and a single repo under the org.

So when a job is generated by the repo, you will receive a webhook on *all* 3 levels and the fields in all of these will be set (repo, org AND enterprise).

At that point, pool managers from all levels will try to allocate a runner for this job. A lock is set on the job by the entity that gets to handle the job first. But (and it's a bit but)! We have no guarantee that the runner we spin up in a pool is the one that will *actually* pick up the job.

So while all 3 fields are set, and....say the `org` entity locked the job and created a runner for it, if the enterprise or the repo has an idle runner already available or will spin up an idle runner before the `org`, then *that* runner will probably pick up the job.

At that point, the job will have a FK set to the runner that actually is handling the job. And that is your real indication of which pool is handling which job in reality.

Scalesets do not suffer from this heuristic, as scheduling is done server side. But pools are a concept internal to GARM, built in a more....dark time.

runner/metrics/jobs.go
```diff
@@ -48,5 +61,78 @@ func CollectJobMetric(ctx context.Context, r *runner.Runner) error {
...
+		}
+
+		key := countKey{
+			Status:     job.Status,
+			EntityType: entityType,
+			EntityName: entityName,
+		}
+		counts[key]++
+
+		// Count scaleset jobs by scaleset name. The scaleset name is Labels[0].
```
gabriel-samfira, line 108: Scalesets have custom labels. You may have more than 1 label for a scaleset, and the first one is *not* guaranteed to be the scaleset name. At least not that I am aware of.

It is perfectly valid for you to define a scaleset called `my-awesome-scaleset` with a custom label called `ubuntu-mem64gb-cpus16` which you can use in `runs-on`.

runner/metrics/metrics.go
```diff
@@ -101,14 +101,14 @@ func collectMetrics(ctx context.Context, r *runner.Runner, controllerInfo params
 		return err
 	}
 
-	slog.DebugContext(ctx, "collecting health metrics")
```
gabriel-samfira, line 104: this reorder has no functional change. Just grows the diff.

runner/metrics/scaleset.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 Cloudbase Solutions SRL
+// Copyright 2025 Cloudbase Solutions SRL
```
gabriel-samfira, line 1: Going back in time? :D

gabriel-samfira (changes requested):
So, most of the low cardinality changes here can be achieved by defining some Prometheus "recording rules". I can set up a gist or add some sample `rules` to the contrib folder if that would help.

New metrics must be additive. Removing or redefining labels will break existing dashboards.

We can keep `garm_scaleset_job_count` and `garm_scaleset_runner_count` if we key them by scale-set `id` (the GARM internal one). That is the only globally unique identifier of a scaleset we manage.

The metrics you removed/re-keyed must be reverted.

